PR #3, opened 2019-01-08
EC2: Refactor delete_user and delete_key functions

asmorodskyi:


webui/ec2/EC2.py
```diff
@@ -78,15 +78,13 @@ def delete_key(key_id):
...
-        return True
+        return delete_user(user.name)
     else:
         iam_res = boto3.resource(
             'iam', aws_access_key_id=provider_conf.EC2['key'],
             aws_secret_access_key=provider_conf.EC2['secret'])
         key = iam_res.AccessKey(user.name, key_id)
-        key.delete()
-        return True
+        return key.delete()
```
cfconrad: If I'm not wrong `delete()` doesn't return something
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iam.html#IAM.AccessKey.delete

webui/ec2/EC2.py
```diff
@@ -98,18 +96,18 @@ def delete_user(username):
...
-            UserName=username):
-        for policy in response.get('AttachedPolicies'):
-            user.detach_policy(PolicyArn=policy.get('PolicyArn'))
-    user.delete()
-    return True
+        for response in iam.get_paginator('list_attached_user_policies').paginate(
+                UserName=username):
+            for policy in response.get('AttachedPolicies'):
+                user.detach_policy(PolicyArn=policy.get('PolicyArn'))
+        return user.delete()
```
cfconrad: same as above.
But the exception handling looks good to me!

PR #7, opened 2019-01-21
Store seen instances in DB and make them visible

cfconrad:
Get instances from EC2 and store them in the local DB.
The local db can be synced with url 'ec2/update'.
To view all instances use 'ec2/instances'.

We use django_filters and django_table2 to create and filter the output.
We use django_bootstrap3 to make the output nice formatted.

@asmorodskyi @jlausuch WDYT?


webui/ec2/EC2.py
```diff
@@ -2,6 +2,21 @@
 from .models import User, AccessKey
 import boto3
 from botocore.exceptions import ClientError
+from django.db import transaction
+from django.utils import timezone
+from ec2.models import Instance
+import json
+
+
+regions = [
```
asmorodskyi: can't you get this list at runtime from Amazon ? Suspecting that this list static more or less so not a big deal actually but still ...

webui/ec2/EC2.py
```diff
@@ -130,3 +145,62 @@ def create_user(username):
     u = User(user.user_name, user.user_id, user.create_date,
              [AccessKey(key.id, key.status, key.create_date, key.secret)])
     return u
+
+
+def list_instances(region='eu-central-1'):
+    ec2 = boto3.resource('ec2', aws_access_key_id=provider_conf.EC2['key'],
```
asmorodskyi: you doing this resource init in almost every function , lack of python knowledge stops me from give you certain suggestions in python but from theoretical OOP perspective I would move this resource in some static field/class and will init once for whole app or at least create ConnectionFactory class which handle all things related to task "give me valid API client instance"

webui/ec2/EC2.py
```diff
@@ -130,3 +145,62 @@ def create_user(username):
...
+                         aws_secret_access_key=provider_conf.EC2['secret'],
+                         region_name=region)
+    return ec2.instances.all()
+
+
+def create_csp_info(i):
```
asmorodskyi: name is misleading , you not creating something here  you just converting boto3 to json, feel free to come up with different one : 
```suggestion
def boto3_to_json(boto3_obj):
```

webui/ec2/EC2.py
cfconrad: I will give it a try.

webui/ec2/EC2.py
cfconrad: agree. Lets do it in a other PR.

webui/ec2/templates/ec2/instance_list.html
```diff
@@ -0,0 +1,24 @@
+{% load render_table from django_tables2 %}
+{% load bootstrap3 %}
+{% load static %}
+<!doctype html>
+<html>
+    <head>
+        <title>List of persons</title>
```
asmorodskyi: maybe list of instances ?

cfconrad: sure, c&p ...

PR #8, opened 2019-01-24
Create EC2 singleton and extract CSP config in file

cfconrad:
Use a singleton for EC2 connection. This avoid creation of multiple
boto3 resources from the same type.

Introduce /ect/pcw.ini file for CSP configuration.

### DIFF
--- a/webui/ec2/views.py
+++ b/webui/ec2/views.py
@@ -1,14 +1,14 @@
 from django.http import HttpResponseForbidden, HttpResponseServerError
 from django.http import HttpResponseNotFound
 from django.shortcuts import redirect
+from webui.settings import ConfigFile
 from rest_framework.response import Response
 from ec2.serializers import UserSerializer, AccessKeySerializer
 from rest_framework import views
 from rest_framework.authentication import SessionAuthentication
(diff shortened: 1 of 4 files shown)


cfconrad:
@asmorodskyi

webui/ec2/views.py
```diff
@@ -23,25 +23,24 @@ class UserView (views.APIView):
...
     def delete(self, request, name):
         # Allow only tmp users to be deleted
-        p = re.compile('^{}$'.format(
-            provider_conf.EC2['username_pattern'].replace('{}', '\\d+')))
+        p = re.compile('^{}$'.format(ConfigFile().get(['default', 'username_pattern']).replace('{}', '\\d+')))
         if not p.match(name):
             return HttpResponseForbidden()
 
-        if EC2.delete_user(name):
+        if EC2().delete_user(name):
```
asmorodskyi: so just for my understanding before we were using here some kind of static method of a class and now we creating real class instance to be able to call method ?

asmorodskyi:
question is still relevant but it more educational one :)

PR #10, opened 2019-03-27
Raise exception, if vault login failed

cfconrad:
This will raise a exception, if the vault credentials are invalid.

### DIFF
--- a/webui/ocw/lib/vault.py
+++ b/webui/ocw/lib/vault.py
@@ -33,13 +33,17 @@ class Vault:
 
     def getClientToken(self):
         login_path = '/v1/auth/userpass/login'
-        if self.client_token is not None:
-            return self.client_token
+        if self.client_token is None:
+            try:
+                r = requests.post(self.url+login_path+'/'+self.user,
+                                  json={'password': self.password},
+                                  verify=self.certificate_dir)
+                if 'errors' in r.json():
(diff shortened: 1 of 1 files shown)


webui/ocw/lib/vault.py
```diff
@@ -33,13 +33,17 @@ def revoke(self):
...
-            return self.client_token
+        if self.client_token is None:
+            try:
+                r = requests.post(self.url+login_path+'/'+self.user,
+                                  json={'password': self.password},
+                                  verify=self.certificate_dir)
+                if 'errors' in r.json():
+                    raise ConnectionError(",".join(r.json()['errors']))
+                self.client_token = r.json()['auth']['client_token']
+            except Exception as e:
```
asmorodskyi: catching exception make sense if you want to mute it , do some recovery and go on . But catching **all** exceptions and rethrow no matter what happened as `ConnectionError` not look nice. For example if for whatever reason you will get broken json or not json at all you will still get ConnectionError which would be confusing. Why not just remove this catch and allow python to do the job ?

cfconrad: As you remember, the exceptions which appear wasn't realy meanfull. And it was hard to understand what actually cause this problem.
Now at least you know that it comes from "Vault Login" and you might check your vault-credentials.

PR #12, opened 2019-03-27
Run update in thread and use ajax to poll status

cfconrad:
As the update takes quite long. We moved this into an background thread.
Then using ajax to poll the status and display it on the list page.

I  continued the work on this topic, to have also a better responding webui on delete.
For this, I introduced a state variable for `Instance`. This can be used to mark `Instance`'s as deleting.
Also add some improvements of the HTML view.


asmorodskyi:
please remove jquery lib file from commit

webui/ocw/views.py
```diff
@@ -54,4 +59,6 @@ def delete(request, key_id=None):
         raise NotImplementedError(
```
asmorodskyi: I can't comment on lines above but you can update lines where you comparing ProviderChoice with your new changes 
if (o.provider == ProviderChoice.AZURE):

PR #13, opened 2019-03-29
Use lease_duration to renew credentials from Vault

cfconrad:
The credentials given from vault have a livetime.
If this expire we need to renew our object.


webui/ocw/lib/azure.py
```diff
@@ -27,22 +27,25 @@ def subscription(self):
         return self.__credentials.getData('subscription_id')
 
     def sp_credentials(self):
```
asmorodskyi: please rename it to CheckCredentials to align with azure

PR #14, opened 2019-03-29
Add email notification for left overs

cfconrad:
The configuration for notification is done in pcw.ini section [notify].
If the system founds an new left-over, it sends an email. The email
contains all current left-overs.


webui/ocw/lib/emailnotify.py
```diff
@@ -0,0 +1,58 @@
...
+            i.instance_id,
+            j['tags']['openqa_created_by'],
+            i.age_formated(),
+            request.build_absolute_uri(reverse(views.delete, args=[i.id]))
+            ])
+
+    if num_new == 0:
+        return
+
+    smtp_server = cfg.get(['notify', 'smtp'])
```
asmorodskyi: if config does not contain such param at all , will we get exception here ?

webui/ocw/views.py
```diff
@@ -45,6 +47,12 @@ def update_status(request):
     return redirect('instances')
 
 
+def cron(request):
```
asmorodskyi: I would say it should require login

webui/ocw/views.py
```diff
@@ -45,6 +47,12 @@ def update_status(request):
...
+    update(request)
```
asmorodskyi: don't think it is good idea we have separate call for that , let's leave it like this

webui/ocw/views.py
```diff
@@ -45,6 +47,12 @@ def update_status(request):
...
+    emailnotify.send_mail(request)
+    return HttpResponse('Done...')
```
asmorodskyi: in case config is empty or something goes wrong we still sending done , right ?

webui/ocw/views.py
asmorodskyi: So will something periodically trigger this function or it is just confusing name ? In second case suggesting to rename it because I am expecting that something with such name should have auto-triggering infra

webui/ocw/views.py
cfconrad: Hm don't get the point. The update() still takes place. And if a email was send or not, is also not visible. So what do you expect here.
For me done means, no exception occurred :)

We could redirect to `/instances` but this would just increase the traffic.

PR #15, opened 2019-04-01
List GCE instances and store in DB

cfconrad:
This is the first step to get GCE support.
It works to list instances and store them into the DB.


webui/ocw/lib/gce.py
```diff
@@ -0,0 +1,65 @@
...
+        if GCE.__instance is None:
+            GCE.__instance = object.__new__(cls)
+            GCE.__instance.__credentials = GCECredential()
+        return GCE.__instance
+
+    def compute_client(self):
+        if self.__credentials.isExpired():
+            self.__credentials.renew()
+            self.__compute_clinet = None
+        self.project = self.__credentials.getInfo()['project_id']
```
asmorodskyi: why project is not listed in the very beginning of the class and don't have '__' before name ?

webui/ocw/lib/vault.py
```diff
@@ -113,16 +114,31 @@ def getCredentials(self):
...
-        cred_file = json.loads(base64.b64decode(self.getData('private_key_data')))
+        if 'errors' in creds:
+            raise Exception(",".join(creds['errors']))
+        cred_file = json.loads(base64.b64decode(creds['data']['private_key_data']).decode(encoding='UTF-8'))
         for k, v in cred_file.items():
             if k not in creds['data']:
                 creds['data'][k] = v
         return creds
+
+    def getInfo(self):
```
asmorodskyi: let's name function according what they do
```suggestion
    def getPrivateKeyData(self):
```

webui/ocw/lib/gce.py
cfconrad: I think, python don't require it.
But I think it make sense. I will change it.

webui/ocw/lib/vault.py
cfconrad: ok

webui/ocw/lib/gce.py
asmorodskyi: yeah I know , just wanted to understand why you treat project field differently from others ?

cfconrad: lazy python programmer :)

PR #17, opened 2019-04-01
vault: simplify constructor and make httpPost generic

asmorodskyi:
1. don't see why and where we could use programmatic way of init Valut sub-classes 
2. try/catch/raise should be used for all http post not only to one which getting token

### DIFF
--- a/webui/ocw/lib/vault.py
+++ b/webui/ocw/lib/vault.py
@@ -12,17 +12,12 @@ class Vault:
     auth_json = None
     auth_expire = None
 
-    def __init__(self, url, user, password, certificate_dir):
-        self.url = url
-        self.user = user
-        self.password = password
-        self.certificate_dir = certificate_dir
+    def __init__(self):
         cfg = ConfigFile()
-        if (self.url is None):
-            self.url = cfg.get(['vault', 'url'])
-            self.user = cfg.get(['vault', 'user'])
-            self.password = cfg.get(['vault', 'password'])
-            self.certificate_dir = cfg.get(['vault', 'cert_dir'], '/etc/ssl/certs')
+        self.url = cfg.get(['vault', 'url'])
+        self.user = cfg.get(['vault', 'user'])
+        self.password = cfg.get(['vault', 'password'])
+        self.certificate_dir = cfg.get(['vault', 'cert_dir'], '/etc/ssl/certs')
 
     def __del__(self):
         self.revoke()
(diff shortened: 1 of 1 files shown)


webui/ocw/lib/vault.py
```diff
@@ -11,53 +11,42 @@ class Vault:
...
-                r = requests.post(self.url+login_path+'/'+self.user,
-                                  json={'password': self.password},
-                                  verify=self.certificate_dir)
-                if 'errors' in r.json():
-                    raise ConnectionError(",".join(r.json()['errors']))
-                self.client_token = r.json()['auth']['client_token']
-            except Exception as e:
-                raise ConnectionError('Vault login failed - {}: {}'.format(type(e).__name__, str(e)))
-
+            r = self.httpPost('/v1/auth/userpass/login'+self.user, data={'password': self.password})
```
cfconrad:
```suggestion
            r = self.httpPost('/v1/auth/userpass/login/'+self.user, data={'password': self.password})
```

webui/ocw/lib/vault.py
```diff
@@ -11,53 +11,42 @@ class Vault:
...
+            self.client_token = {'X-Vault-Token': r.json()['auth']['client_token']}
```
cfconrad: This will break `httpGet()`. The headers should be build differently in `httpGet()` now.

webui/ocw/lib/vault.py
```diff
@@ -84,8 +73,8 @@ class AzureCredential(Vault):
     ''' Known data fields: subscription_id, client_id, client_secret, tenant_id
     '''
 
-    def __init__(self, url=None, user=None, password=None, certificate_dir=None):
-        return super().__init__(url, user, password, certificate_dir)
+    def __init__(self):
```
cfconrad: As long as we do not need to do something in that initializer, we don't need to have it at all! And then the super() initializer is called anyway.

webui/ocw/lib/vault.py
```diff
@@ -102,8 +91,8 @@ def getCredentials(self):
 class EC2Credential(Vault):
     ''' Known data fields: access_key, secret_key '''
 
-    def __init__(self, url=None, user=None, password=None, certificate_dir=None):
-        return super().__init__(url, user, password, certificate_dir)
+    def __init__():
```
cfconrad: see above

webui/ocw/lib/vault.py
```diff
@@ -114,8 +103,8 @@ class GCECredential(Vault):
     ''' Known data fields: private_key_data, project_id, private_key_id,
             private_key, client_email, client_id'''
 
-    def __init__(self, url=None, user=None, password=None, certificate_dir=None):
-        return super().__init__(url, user, password, certificate_dir)
+    def __init__():
```
cfconrad: see above

webui/ocw/lib/vault.py
```diff
@@ -11,53 +11,42 @@ class Vault:
...
 
-    def httpPost(self, path, data):
-        return requests.post(self.url + path,
-                             json=data,
-                             headers={'X-Vault-Token': self.getClientToken()},
-                             verify=self.certificate_dir)
+    def httpPost(self, path, data, headers={}):
+        try:
+            r = requests.post(self.url + path, json=data, headers=headers, verify=self.certificate_dir)
+            if 'errors' in r.json():
```
cfconrad: As now the error handling for Post looks generic and this function will not raise something different then ConnectionError() we should do the same with httpGet() or?

asmorodskyi: yes but I would like to do it in different PR if it is ok for you

cfconrad: ok, but you could just add the same error handling for both?

webui/ocw/lib/vault.py
asmorodskyi: I was not sure how constructors works in python so decided to leave it . Now I will remove it

webui/ocw/lib/vault.py
asmorodskyi: done

webui/ocw/lib/vault.py
asmorodskyi: fixed

webui/ocw/lib/vault.py
asmorodskyi: done

webui/ocw/lib/vault.py
```diff
@@ -11,53 +11,49 @@ class Vault:
...
 
     def __del__(self):
         self.revoke()
 
     def revoke(self):
         if self.auth_json is None:
             return
-        path = '/v1/sys/leases/revoke'
-        self.httpPost(path, {'lease_id': self.auth_json['lease_id']})
+        self.httpPost('/v1/sys/leases/revoke', {'lease_id': self.auth_json['lease_id']}, self.getClientToken)
```
cfconrad:
```suggestion
        self.httpPost('/v1/sys/leases/revoke', {'lease_id': self.auth_json['lease_id']}, self.getClientToken())
```

asmorodskyi:
work fine

PR #18, opened 2019-04-01
Vault: refactor credentials

asmorodskyi:


asmorodskyi:
> I don't like the idea of holding some CSP state object and Vault-connection logic inside of one class.

Can you give some reasoning ? 

> What do you think on inherit from the Vault XXXCredentials class in EC2 and Azure and override the `revoke()` message to cleanup the CSP states, if the credentials changed.

sorry not sure that I get fully your idea . Let's discuss it tomorrow . 

Generally speaking I don't like this version also . Just like it more then master version. So if this will be accepted I plan to refactor one more time :) at the end I see drop of all credentials classes and move all logic back to CSP classes but together with getcredentials. The fate of Vault class is not clear ... also probably we need some parent class for CSP classes ...
Let's speak tomorrow :)

cfconrad:
> Can you give some reasoning ?

The reason is, that I think the communication with Vault is an isolated task. And also later easy to test, if this class keept small and simple.

Yes lets have a talk later. I'm back at 11am, ping you then.

PR #19, opened 2019-04-02
Inherit from Vault for the provider classes.

cfconrad:


cfconrad:
there will be a better solution...

PR #21, opened 2019-04-03
refactor isExpired usage

asmorodskyi:


webui/ocw/lib/EC2.py
```diff
@@ -39,27 +27,32 @@ def check_credentials(self):
                 time.sleep(1)
         raise Exception("Invalid EC2 credentials")
 
-    def ec2_resource(self, region='eu-central-1'):
-        if region not in self.__ec2_resource:
-            self.__ec2_resource[region] = boto3.resource('ec2', aws_access_key_id=self.__key,
-                                                         aws_secret_access_key=self.__secret,
+    def get_resource_region(self, region='eu-central-1'):
+        if region not in self.__resource_regions or self.__credentials.isExpired():
+            key, secret = self.getKey_and_secret()
```
cfconrad: I think it will not work. The `isExpired()` will return only once `True` and then you are calling `getData()` which lead to new credentials and new expire date.
A call later call to `get_client_region()` will not recreate the `self.__client_regions` with new key and pw, or?

webui/ocw/lib/EC2.py
```diff
@@ -39,27 +27,32 @@ def check_credentials(self):
...
     def list_instances(self, region='eu-central-1'):
-        return self.ec2_resource(region).instances.all()
+        return self.get_resource_region(region).instances.all()
 
     def list_regions(self):
-        regions_resp = self.ec2_client().describe_regions()
-        regions = [region['RegionName'] for region in regions_resp['Regions']]
-        return regions
+        regions_resp = self.get_client_region().describe_regions()
+        self.__client_regions = [region['RegionName'] for region in regions_resp['Regions']]
```
cfconrad: don't know, whay you overwrite the `self.__client_regions` here.
When I understood your change correct, the dict `self__client_regions` is a dict which holds a boto3.client(ec2) object per region.
This `list_regions()` method, is to retrieve all supported regions by given credentials. So a list of strings.

webui/ocw/lib/EC2.py
```diff
@@ -5,10 +5,8 @@
 
 class EC2:
     __instance = None
-    __key = None
-    __secret = None
-    __ec2_resource = dict()
-    __ec2_client = dict()
+    __resource_regions = dict()
+    __client_regions = dict()
```
cfconrad: maybe:
```suggestion
    __client_by_region = dict()
```

webui/ocw/lib/EC2.py
```diff
@@ -5,10 +5,8 @@
 
 class EC2:
     __instance = None
-    __key = None
-    __secret = None
-    __ec2_resource = dict()
-    __ec2_client = dict()
+    __resource_regions = dict()
```
cfconrad: maybe:
```suggestion
    __resource_by_region = dict()
```

webui/ocw/lib/azure.py
```diff
@@ -44,21 +40,21 @@ def check_credentials(self):
         raise AuthenticationError("Invalid Azure credentials")
 
     def sp_credentials(self):
-        if (self.__sp_credentials is None):
+        if (self.__sp_credentials is None or self.__credentials.isExpired()):
```
cfconrad: Same problem as in EC2(). The `self.__compute_mgmt_client` and `self.__resource_mgmt_client` will not renewed.

webui/ocw/lib/EC2.py
asmorodskyi: done

webui/ocw/lib/EC2.py
asmorodskyi: done

PR #23, opened 2019-04-04
List instances in GCE from all zones

cfconrad:
Like on EC2, GCE has multiple regions with multiple zones. So we need to iterate over each zone and list the instances.

### DIFF
--- a/webui/ocw/lib/EC2db.py
+++ b/webui/ocw/lib/EC2db.py
@@ -35,14 +35,13 @@ def _instance_to_json(i):
 @transaction.atomic
 def sync_instances_db(region, instances):
     o = Instance.objects
-    o = o.filter(region=region, provider=ProviderChoice.EC2, state=StateChoice.ACTIVE)
-    o = o.update(state=StateChoice.UNK, active=False)
+    o = o.filter(region=region, provider=ProviderChoice.EC2)
+    o = o.update(active=False)
 
     for i in instances:
         db.update_or_create_instance(
                 provider=ProviderChoice.EC2,
                 instance_id=i.instance_id,
-                active=i.state['Name'] != 'terminated',
                 region=region,
                 csp_info=_instance_to_json(i))
 
--- a/webui/ocw/lib/azure.py
+++ b/webui/ocw/lib/azure.py
@@ -64,10 +64,10 @@ class Azure:
         return self.__resoure_mgmt_client
 
     def list_instances(self):
-        return self.compute_mgmt_client().virtual_machines.list_all()
+        return [i for i in self.compute_mgmt_client().virtual_machines.list_all()]
 
     def list_resource_groups(self):
-        return self.resource_mgmt_client().resource_groups.list()
+        return [r for r in self.resource_mgmt_client().resource_groups.list()]
 
     def delete_resource(self, resource_id):
         return self.resource_mgmt_client().resource_groups.delete(resource_id)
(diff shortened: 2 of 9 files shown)


webui/ocw/lib/EC2db.py
```diff
@@ -35,8 +35,8 @@ def _instance_to_json(i):
 @transaction.atomic
 def sync_instances_db(region, instances):
     o = Instance.objects
-    o = o.filter(region=region, provider=ProviderChoice.EC2, state=StateChoice.ACTIVE)
-    o = o.update(state=StateChoice.UNK, active=False)
+    o = o.filter(region=region, provider=ProviderChoice.EC2)
```
asmorodskyi: how this relates to GCE ? if not can you please explain what is happening here ?

webui/ocw/lib/azure.py
```diff
@@ -89,8 +89,8 @@ def _instance_to_json(i):
 @transaction.atomic
 def sync_instances_db(instances):
     o = Instance.objects
-    o = o.filter(provider=ProviderChoice.AZURE, state=StateChoice.ACTIVE)
-    o = o.update(active=False, state=StateChoice.UNK)
+    o = o.filter(provider=ProviderChoice.AZURE)
```
asmorodskyi: same questions , see above

webui/ocw/lib/EC2db.py
cfconrad: This is really not directly related to GCE, but I figured that out during GCE extending.
The thing is that the synchronization between local DB and the CSP status was a bit obfuscated.

It works now like:
  * Instance.active is used to identify if such entry exists on CSP
  * Instance.state is used to reflect the latest "local" state, e.g. if delete button was pressed.

asmorodskyi: can you please add this to field description ?

asmorodskyi: as comments where you declare Instance ( db.py )

webui/ocw/lib/EC2db.py
```diff
@@ -35,8 +35,8 @@ def _instance_to_json(i):
...
+    o = o.update(active=False)
```
asmorodskyi: and why it's getting active=False here ?

webui/ocw/lib/db.py
```diff
@@ -21,7 +21,7 @@ def update_or_create_instance(provider, instance_id, region, csp_info, active=Tr
         o.age = t_now - o.first_seen
         o.active = active
         o.csp_info = json.dumps(csp_info, ensure_ascii=False)
-        if o.state == StateChoice.UNK:
+        if o.state != StateChoice.DELETING:
```
asmorodskyi: and if it's DELETED ? 
and if it is DELETING why we saving the state ?

webui/ocw/lib/gce.py
```diff
@@ -30,35 +30,75 @@ def compute_client(self):
...
 
-    def list_instances(self, zone='europe-west1-b'):
-        i = self.compute_client().instances().list(project=self.__project, zone=zone).execute()
-        return i['items'] if 'items' in i else []
+    def list_instances(self, zone):
+        result = []
+        request = self.compute_client().instances().list(project=self.__project, zone=zone)
+        while request is not None:
+            response = request.execute()
+            if 'items' in response:
```
asmorodskyi: and if there is no 'items' in response ? maybe we need it least log this ?

webui/ocw/lib/gce.py
```diff
@@ -30,35 +30,75 @@ def compute_client(self):
             self.__compute_clinet = googleapiclient.discovery.build('compute', 'v1', credentials=credentials)
         return self.__compute_clinet
 
-    def list_instances(self, zone='europe-west1-b'):
-        i = self.compute_client().instances().list(project=self.__project, zone=zone).execute()
-        return i['items'] if 'items' in i else []
+    def list_instances(self, zone):
```
asmorodskyi: what method is doing looks strange and requires some explanations ( even for yourself are sure that you won't  forget what's happening here in 6 months ? )

webui/ocw/lib/gce.py
cfconrad: Don't know. Thins only means, that there  is no instance running at CSP. Which could happen.

webui/ocw/lib/db.py
cfconrad: Add a  warning, if we change from DELETED to something else

asmorodskyi:
LGTM , is it ready to merge ?

PR #26, opened 2019-04-23
Boto3 image chk meta

cfconrad:
Show exception on failures in Update.

Fix: lazy boto3 Image object access
If an Image doesn't exists anymore, the Image.meta.data is None.
So an access to Image.name raise an Exception.

@asmorodskyi

### DIFF
--- a/webui/ocw/lib/db.py
+++ b/webui/ocw/lib/db.py
@@ -6,9 +6,13 @@ import json
 import dateutil.parser
 from threading import Thread
 from threading import Lock
+from datetime import datetime
+from .emailnotify import send_mail
+import traceback
 import logging
 
 update_thread = None
+update_date = None
 update_mutex = Lock()
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


webui/ocw/lib/db.py
```diff
@@ -54,47 +55,53 @@ def __update_run():
...
 
     '''
     Each update is using Instance.active to mark the model is still availalbe on CSP.
     Instance.state is used to reflect the "local" state, e.g. if someone triggered a delete, the
     state will moved to DELETING. If the instance is gone from CSP, the state will set to DELETED.
     '''
+    try:
+        instances = Azure().list_resource_groups()
+        logger.info('Got {} resources groups from Azure'.format(len(instances)))
+        azure.sync_instances_db(instances)
```
asmorodskyi: maybe better to add this try/catch thingy into each provider ? this patch for sure an improvement into comparison  what we had before ( we getting at least some error handling ) but still - you don't want to fail Azure DB update when you having troubles with Amazon , right ?

cfconrad: hm, but doing it in each provider, would need a additional mechanize to announce the errors back to this update block.
I clould do a single try/catch per provider.

cfconrad: Thinking twice on it. I actually need to investigate in any case, if there is an exception on update().
Event if it isn't that nice, but I could stop continue.
We could think on sending an email.

asmorodskyi: as I said this version is already improvement , so it's actually fine. But imagine you have issue with update Azure and Amazon simultaneously :) with current code you will be not even aware about issue with Amazon update before you will not fix Azure issue :)  again I am ready to merge as it is because it would be improvement anyway .
Sending emails would be cool .  Do you want to do it in this PR or separate one ?

cfconrad: Agree, lets do it in a separate one. I think I would then go with a  single thread per provider. So we also have a time improvement on updating.

webui/ocw/lib/db.py
```diff
@@ -54,47 +55,53 @@ def __update_run():
...
-        EC2db.sync_instances_db(region, instances)
-
-    instances = GCE().list_all_instances()
-    logger.info('Got {} instances from GCE'.format(len(instances)))
-    gce.sync_instances_db(instances)
+        instances = GCE().list_all_instances()
+        logger.info('Got {} instances from GCE'.format(len(instances)))
+        gce.sync_instances_db(instances)
+    except Exception as e:
+        logger.exception('Unexpected error on update instances')
```
asmorodskyi: why you not writing `e` content into a log ?

webui/ocw/lib/db.py
```diff
@@ -54,47 +55,53 @@ def __update_run():
     from . import EC2db
     from . import azure
     from . import gce
+    global update_error
```
asmorodskyi: don't like the idea of this variable . it keeping info about error only till next update round. this give a room for different messy scenarios . Like one time error , sporadic error , different errors with each run. I would drop it at all and setup mailing of errors . it would be good enough and is_updating function should just answer if update is currently running or not. Probably would be nice also add time of last successful update

asmorodskyi: also if you drop it  this will ease implementation of my idea to moving try/catch into providers functions :)

webui/ocw/lib/db.py
cfconrad: logger.exception() does it. https://docs.python.org/3/library/logging.html#logging.Logger.exception

asmorodskyi: ah ok

webui/ocw/lib/db.py
cfconrad: Ok, I will send a email notification when an error occur. But also display something in  the GUI. E.g if sending mail fail, at least in the GUI is something visible.

asmorodskyi: only in case you are lucky to check UI before next iteration :) what about my idea to just display time of last successful update ? if it is too old you will understand that something is wrong and go to check the logs to see what is actually wrong

cfconrad: done

PR #27, opened 2019-04-25
Add config option ec2.regions

cfconrad:
Regions can be configured per account.
https://console.aws.amazon.com/billing/home?#/account?AWS-Regions
So we need to specify a list of regions manually.

### DIFF
--- a/webui/ocw/lib/db.py
+++ b/webui/ocw/lib/db.py
@@ -1,3 +1,4 @@
+from webui.settings import ConfigFile
 from ..models import Instance
 from ..models import StateChoice
 from django.db import transaction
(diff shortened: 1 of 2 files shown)


webui/ocw/lib/db.py
```diff
@@ -70,7 +72,7 @@ def __update_run():
         logger.info('Got {} resources groups from Azure'.format(len(instances)))
         azure.sync_instances_db(instances)
 
-        for region in EC2().list_regions():
+        for region in cfg.getList(['ec2', 'regions'], EC2().list_regions()):
```
asmorodskyi: not 100% sure how it is done in python , but maybe better move getList out of the loop so we will not call `EC2().list_regions()` on every iteration ?

PR #30, opened 2019-04-29
emailnotify: add exception name to subject

asmorodskyi:


webui/ocw/lib/emailnotify.py
```diff
@@ -63,7 +63,7 @@ def send_mail(subject, message):
 
 {message}
 '''.format(subject=subject, _from=sender_email, _to=receiver_email, message=message)
-    logger.info("Send Email To:'{}' Subject:'{}'".format(receiver_email, subject))
+    logger.info("Send Email To:'{}' Subject:'[Openqa-Cloud-Watch] {}'".format(receiver_email, subject))
```
cfconrad: Lets add this to line #59 too

asmorodskyi:
I think you know my answer :) - in current implementation we need it always when we will have case where we will need to not have it or have different one we can refactor code to support it :)

PR #34, opened 2019-09-05
Add vault-namespace when requesting credentials

cfconrad:
Adding support for vault-namespace concept to collect VM's from
different subscriptions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ cat > /etc/pcw.ini << EOT
 url = https://publiccloud.your.vault.server/vault
 user = Your_VAULT_USER
 password = VAULT_USER_PASSWORD
+namespaces = VAULT_NAMESPACE, VAULT_NAMESPACE_2, ...
+
+[ec2]
+regions = eu-north-1, ap-south-1, eu-west-3, ...
 
 # Add this section to enable email notification for left overs
 [notify]
--- a/README.md
+++ b/README.md
@@ -39,6 +39,11 @@ user = Your_VAULT_USER
 password = VAULT_USER_PASSWORD
 namespaces = VAULT_NAMESPACE, VAULT_NAMESPACE_2, ...
 
+[vault.namespace.XXX]
+# XXX is the name of a namespace given in vault.namespaces
+# provider should be ec2, azure or csp
+providers = csp1[, csp2]...
+
 [ec2]
 regions = eu-north-1, ap-south-1, eu-west-3, ...
 
--- a/webui/ocw/lib/EC2.py
+++ b/webui/ocw/lib/EC2.py
@@ -5,7 +5,7 @@ import logging
 
 
 class EC2:
-    __instance = None
+    __instances = dict()
     __key = None
     __secret = None
     __ec2_resource = dict()
@@ -13,14 +13,14 @@ class EC2:
     __credentials = None
     __logging = None
 
-    def __new__(cls):
-        if EC2.__instance is None:
-            EC2.__instance = object.__new__(cls)
-            EC2.__instance.__credentials = EC2Credential()
-            EC2.__instance.__logging = logging.getLogger(__name__)
+    def __new__(cls, vault_namespace):
+        if vault_namespace not in EC2.__instances:
+            EC2.__instances[vault_namespace] = object.__new__(cls)
+            EC2.__instances[vault_namespace].__credentials = EC2Credential(vault_namespace)
+            EC2.__instances[vault_namespace].__logging = logging.getLogger(__name__)
 
-        EC2.__instance.check_credentials()
-        return EC2.__instance
+        EC2.__instances[vault_namespace].check_credentials()
+        return EC2.__instances[vault_namespace]
 
     def check_credentials(self):
         if self.__credentials.isExpired():
--- a/webui/ocw/lib/EC2db.py
+++ b/webui/ocw/lib/EC2db.py
@@ -35,9 +35,9 @@ def _instance_to_json(i):
 
 
 @transaction.atomic
-def sync_instances_db(region, instances):
+def sync_instances_db(region, instances, vault_namespace):
     o = Instance.objects
-    o = o.filter(region=region, provider=ProviderChoice.EC2)
+    o = o.filter(region=region, provider=ProviderChoice.EC2, vault_namespace=vault_namespace)
     o = o.update(active=False)
 
     for i in instances:
@@ -45,7 +45,8 @@ def sync_instances_db(region, instances):
                 provider=ProviderChoice.EC2,
                 instance_id=i.instance_id,
                 region=region,
-                csp_info=_instance_to_json(i))
+                csp_info=_instance_to_json(i),
+                vault_namespace=vault_namespace)
 
     o = Instance.objects
     o = o.filter(region=region, provider=ProviderChoice.EC2, active=False)
--- a/webui/ocw/lib/azure.py
+++ b/webui/ocw/lib/azure.py
@@ -13,21 +13,21 @@ import logging
 
 
 class Azure:
-    __instance = None
+    __instances = dict()
     __credentials = None
     __compute_mgmt_client = None
     __sp_credentials = None
     __resource_mgmt_client = None
     __logger = None
 
-    def __new__(cls):
-        if Azure.__instance is None:
-            Azure.__instance = object.__new__(cls)
-            Azure.__instance.__credentials = AzureCredential()
-            Azure.__instance.__logger = logging.getLogger(__name__)
+    def __new__(cls, vault_namespace):
+        if vault_namespace not in Azure.__instances:
+            Azure.__instances[vault_namespace] = object.__new__(cls)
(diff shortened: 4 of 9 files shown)


cfconrad:
@pdostal this should solve the problem to see instances created with different vault namespace.

README.md
```diff
@@ -37,6 +37,15 @@ cat > /etc/pcw.ini << EOT
 url = https://publiccloud.your.vault.server/vault
 user = Your_VAULT_USER
 password = VAULT_USER_PASSWORD
+namespaces = VAULT_NAMESPACE, VAULT_NAMESPACE_2, ...
+
+[vault.namespace.XXX]
+# XXX is the name of a namespace given in vault.namespaces
+# provider should be ec2, azure or csp
+providers = csp1[, csp2]...
```
asmorodskyi: why we need this ? isn't it easier just apply namespace for all providers ?

webui/ocw/lib/EC2.py
```diff
@@ -5,22 +5,22 @@
...
     __ec2_client = dict()
     __credentials = None
     __logging = None
 
-    def __new__(cls):
-        if EC2.__instance is None:
-            EC2.__instance = object.__new__(cls)
-            EC2.__instance.__credentials = EC2Credential()
-            EC2.__instance.__logging = logging.getLogger(__name__)
+    def __new__(cls, vault_namespace):
```
asmorodskyi: something broken here from OOP perspective . Or  at least from Java-like OOP perspective . If we stating that class EC2 is keeping **collection** of all instances new method need to accept collection of namespaces  and return collection of instances. or we need another class e.g.  EC2_collection

webui/ocw/lib/EC2db.py
```diff
@@ -35,17 +35,18 @@ def _instance_to_json(i):
 
 
 @transaction.atomic
-def sync_instances_db(region, instances):
+def sync_instances_db(region, instances, vault_namespace):
```
asmorodskyi: why we not sync all namespaces in one shot ?

asmorodskyi:
I think you try to keep module too flexible "just in case" . This force you to add a lot of dimension to a model which we actually not need . Or I don't see clear use case for them

webui/ocw/lib/EC2db.py
pdostal: You need special credentials for each namespace.

asmorodskyi: how is that related ? if you not modify `sync_instances_db` method to take into account namespace this would lead to not filtering by namespace in line 40 which would lead that loop in line 43 will send to update_or_create_instance **all** available instances of EC2 ( each one still having its own credentials )

README.md
cfconrad: This is the default => and this hint is missing in the here, will do it
see: https://github.com/cfconrad/pcw/pull/34/files#diff-0ad2fc8593cc924b952fd5277e6a11d7R76

We need this, cause currently, not all credentials are well configured. For instance I think `qam` has only configured *ec2*. @pdostal plz correct me if I'm wrong.
So with out it, we will never get successful update.

webui/ocw/lib/EC2.py
cfconrad: I think this is a classical case for a `Factory` with cache, but I was to lazy to implement it and used a `singleton` approach. Actually `singleton` makes also sense here, as we do wont this kind of object only once in the application.
But if you don't like it, I would make a proposal with a `Factory`. Maybe next PR, as this is quite big?

webui/ocw/lib/EC2db.py
cfconrad: The instance doesn't know to which "vault-namespace" it belongs to, thats why I'm adding it here.
We could also hook the get_instances() methods to add this parameter directly into the instance. Or let get_instances() directly return `ocw.models.Instance` objects.

webui/ocw/lib/EC2.py
asmorodskyi: this is my point you can not call this singleton ( again maybe just some weird python/perl singleton but not **that** https://en.wikipedia.org/wiki/Singleton_pattern singelton ) , because : 
1. singleton is not allowing to call `new` , you can get instance of object only via getInstance() 
2. singleton ( as it comes from its name  ) always having **single** instance here you having array of instances

webui/ocw/lib/EC2db.py
asmorodskyi: > The instance doesn't know to which "vault-namespace" it belongs to, thats why I'm adding it here.

if instance don't know "vault-namespace" how this is possible ? : 
`o = o.filter(region=region, provider=ProviderChoice.EC2, vault_namespace=vault_namespace)`

webui/ocw/lib/EC2.py
asmorodskyi: https://www.tutorialspoint.com/python_design_patterns/python_design_patterns_singleton.htm python version of SIngleton .

README.md
pdostal: That is true. We still don't have any other provider than EC2 available.

asmorodskyi: let's leave this to the end , in case patch would be simplified in a way I see it , this will not be an issue that you don't have all configured

webui/ocw/lib/EC2db.py
cfconrad: you need to differ between two types of instance. We have a local one in our DB this is what you see here and the one in the CSP which doesn't know vault-namespaces.

webui/ocw/lib/EC2.py
cfconrad: Then lets call it decorator of a collection from singletons :)

There are some more ways to make singletons in python: https://python-3-patterns-idioms-test.readthedocs.io/en/latest/Singleton.html
And as you are not able to protect a constructor in python, you need othere ways as described in yours or mine link.
Not sure why I used a different one then in yours. Maybe because I thought it doesn't make sense to add a getInstance() call when I'm not able to hide the constructor at all.

Anyway, we need to say if we can keep it or not.

webui/ocw/lib/EC2db.py
asmorodskyi: 1. does not matter , what matters is that this `o` which is result of filter method suppose to know about namespaces, so this give me confidence to say that method sync_instances_db may exists without input parameter `namespaces ` 
2.  didn't get how the one that is in CSP is related to this function ? can you point me to line in repository where we trying to query instance  object **in** CSP ?

webui/ocw/lib/EC2.py
asmorodskyi: > Then lets call it decorator of a collection from singletons :)

this discussion goes too deep into theoretical field :P 

> Anyway, we need to say if we can keep it or not.

totally agree . But let's conclude thread  around `sync_instance_db`  and "useless" namespace there . this would influence a lot in my opinion here ...

webui/ocw/lib/EC2db.py
cfconrad: 1) The problem is that you get a subset of all instances in db from the provider. And this subset has common parameters which differs to other subsets. And one of these keys are vault_namespace.

2.)Sure, it is always a similar pattern.
 **Azure:** 
foreach vault_namespaces do:
   i = Azure(namespace)->[get_instances_from_csp](https://github.com/cfconrad/pcw/pull/34/files#diff-0ad2fc8593cc924b952fd5277e6a11d7R79);
   [sync_instances_with_local_db(i,ns)](https://github.com/cfconrad/pcw/pull/34/files#diff-0ad2fc8593cc924b952fd5277e6a11d7R81)

**Ec2** (It's a bit different, as the instance didn't contain the region, but we need the region to destroy the instance). Thats why it is:
foreach vault namespace
   foreach region
       i = EC2(region, namespace)->[get_instances_from_csp](https://github.com/cfconrad/pcw/pull/34/files#diff-0ad2fc8593cc924b952fd5277e6a11d7R85)
       [sync_instances_with_local_db(i,ns)](https://github.com/cfconrad/pcw/pull/34/files#diff-0ad2fc8593cc924b952fd5277e6a11d7R87)

webui/ocw/lib/EC2db.py
```diff
@@ -35,17 +35,18 @@ def _instance_to_json(i):
...
     o = Instance.objects
```
asmorodskyi:
```suggestion
    allObjects = Instance.objects
```

webui/ocw/lib/EC2db.py
```diff
@@ -35,17 +35,18 @@ def _instance_to_json(i):
...
-    o = o.filter(region=region, provider=ProviderChoice.EC2)
+    o = o.filter(region=region, provider=ProviderChoice.EC2, vault_namespace=vault_namespace)
```
asmorodskyi:
```suggestion
    filteredObjects = allObjects.filter(region=region, provider=ProviderChoice.EC2, vault_namespace=vault_namespace)
```

asmorodskyi: it is really insane to read and discuss this code :(  

>object o which is result of filtering of object o , but it is not same object ....

PR #36, opened 2019-09-11
Collect all DB related logic in one place

asmorodskyi:
Collect all DB related logic in one place


webui/ocw/lib/db.py
```diff
@@ -19,49 +27,111 @@
...
-            csp_info=json.dumps(csp_info, ensure_ascii=False)
-        )
         o.age = o.last_seen - o.first_seen
         o.save()
+    o = Instance.objects
+    o = o.filter(provider=provider, active=False)
+    o = o.update(state=StateChoice.DELETED)
+
+def ec2_to_json(i):
+    # TODO find a generic way from boto3 object to json
```
cfconrad: I doubt we will find a generic way, as all of them are different :)

webui/ocw/lib/db.py
```diff
@@ -19,49 +27,111 @@
 
 
 @transaction.atomic
-def update_or_create_instance(provider, instance_id, region, csp_info, vault_namespace):
+def sync_pcp_to_local_db(pc_instances, provider, vault_namespace):
```
cfconrad:
```suggestion
def sync_csp_to_local_db(pc_instances, provider, vault_namespace):
```

asmorodskyi:
> Nice done!
> Did you run it?

nope , also code actually require some import clean up etc. , but didn't want to invest time into this before I generally confirm with you that you think it is right way to go

webui/ocw/lib/db.py
asmorodskyi: ok , actually I was not sure about name just knew that it should not be the old one :) so any name from is fine for me ... except the old one :D

PR #37, opened 2019-10-09
Include namespace and base URL in emailnotification

cfconrad:
Email now looks like:

```
Subject: [Openqa-Cloud-Watch] CSP left overs
From: pcw@publiccloud.qa.suse.de
To: bla@example.com

Message from http://127.0.0.1:8000/


Provider           id              Created-By     Namespace    Age                Delete             
=====================================================================================================
EC2        i-05e3451a1debb7832   openqa-suse-de   qam         1h28m   http://127.0.0.1:8000/delete/29
```

### DIFF
--- a/webui/ocw/lib/emailnotify.py
+++ b/webui/ocw/lib/emailnotify.py
@@ -23,7 +23,7 @@ def send_leftover_notification(request):
 
     table = Texttable(max_width=0)
     table.set_deco(Texttable.HEADER)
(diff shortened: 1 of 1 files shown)


asmorodskyi:
I would expect more "involvement" of namespace : 
1. Put it into subj so I can filter out namespaces which I am not interested 
2. introduce different lists of recipients  in config

hm actually if you will do second you don't need first :)

asmorodskyi:
but also you consider previous comment as TODO for next round :) 

LGTM ( up to you to merge it or not )

PR #38, opened 2019-10-09
Namespace receipient mail notification

cfconrad:
This allow a specific section in pcw.ini file to set email recipients
for a defined vault namespace.

This PR is on top of https://github.com/cfconrad/pcw/pull/37


cfconrad:
From https://github.com/cfconrad/pcw/pull/37
>     1. Put it into subj so I can filter out namespaces which I am not interested
For this you can get only mails from a specific namespace or all.

>     2. introduce different lists of recipients  in config
this is what it does.

PR #39, opened 2019-10-14
Fix sending out to many notification mails

cfconrad:
There was an issue when searching for not notified Instances.

We need to check, if in our leftover list are existing instances with
field `notified==False` and only then send mails to the corresponding
recipients.


asmorodskyi:
I also have some ideas how to improve this flow ... but this will work for now

PR #40, opened 2019-11-07
Introduce vault response caching for development

cfconrad:
When we get new credentials from vault, it takes some time (depending on
CSP) till these get valid on CSP side. To avoid this waiting period and
avoiding of creation from unnecessary many credentials, you could set the
vault.use_cache_file=True option in pcw.ini.

### DIFF
--- a/webui/ocw/lib/vault.py
+++ b/webui/ocw/lib/vault.py
@@ -4,14 +4,19 @@ import json
 from webui.settings import ConfigFile
 from datetime import datetime
 from datetime import timedelta
+import dateutil.parser
 from tempfile import NamedTemporaryFile
+from pathlib import Path
+import logging
+import os
+
+logger = logging.getLogger(__name__)
 
 
 class Vault:
     client_token = None
     client_token_expire = None
     auth_json = None
-    auth_expire = None
     extra_time = 600
 
     def __init__(self, vault_namespace):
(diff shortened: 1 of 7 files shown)


webui/ocw/lib/vault.py
```diff
@@ -21,6 +26,7 @@ def __init__(self, vault_namespace):
         self.namespace = vault_namespace
         self.password = cfg.get(['vault', 'password'])
         self.certificate_dir = cfg.get(['vault', 'cert_dir'], '/etc/ssl/certs')
+        self.use_file_cache = cfg.getBoolean(['vault', 'use_file_cache'])
```
asmorodskyi: so basically to change this you need to restart server , right ? 
what if we would just check config file every time so basically you can change this setting on the fly ?

cfconrad: We currently read config file on the fly see https://github.com/cfconrad/pcw/blob/master/webui/webui/settings.py#L184

The problem is, it will only be used if you get "new" credentials and for this you would need to wait that 2h. 
But when you change the code during development, django already reload the server. So you get once new credentials and then these are stored into the cache file.

webui/ocw/lib/vault.py
```diff
@@ -90,6 +101,35 @@ def renew(self):
         self.revoke()
         self.getData()
 
+    def getAuthCacheFile(self):
+        oldmask = os.umask(0o077)
```
asmorodskyi: why you need this variable ?

webui/ocw/lib/vault.py
cfconrad: > The problem is, it will only be used if you get "new" credentials and for this you would need to > wait that 2h.
that was wrong, as you correctly mentioned, the constructor where we store it in `self.use_file_cache` is only called once. So yes I will change it.

webui/ocw/lib/vault.py
cfconrad: To restore the umask afterwards

webui/ocw/lib/vault.py
```diff
@@ -90,6 +101,35 @@ def renew(self):
         self.revoke()
         self.getData()
 
+    def getAuthCacheFile(self):
```
asmorodskyi: I think this name would be more accurate 
```suggestion
    def createAuthCachePath(self):
```

webui/ocw/lib/vault.py
```diff
@@ -90,6 +101,35 @@ def renew(self):
...
+        path = Path('/tmp/pcw/{}/{}'.format(self.__class__.__name__, self.namespace))
+        path.mkdir(mode=0o700, parents=True, exist_ok=True)
+        f = path.joinpath('auth.json')
+        f.touch(mode=0o600, exist_ok=True)
+        oldmask = os.umask(oldmask)
+        return f
+
+    def loadFileCache(self):
+        try:
+            filename = self.getAuthCacheFile()
```
asmorodskyi:
```suggestion
            authjson = self.getAuthCacheFile()
```

webui/ocw/lib/vault.py
```diff
@@ -90,6 +101,35 @@ def renew(self):
...
+            with filename.open() as f:
```
asmorodskyi:
```suggestion
            with authjson.open() as authjson_file:
```

webui/ocw/lib/vault.py
```diff
@@ -90,6 +101,35 @@ def renew(self):
...
+    def getAuthCacheFile(self):
+        oldmask = os.umask(0o077)
+        path = Path('/tmp/pcw/{}/{}'.format(self.__class__.__name__, self.namespace))
+        path.mkdir(mode=0o700, parents=True, exist_ok=True)
+        f = path.joinpath('auth.json')
+        f.touch(mode=0o600, exist_ok=True)
+        oldmask = os.umask(oldmask)
+        return f
+
+    def loadFileCache(self):
```
asmorodskyi:
```suggestion
    def loadAuthJson(self):
```

webui/ocw/lib/vault.py
```diff
@@ -90,6 +101,35 @@ def renew(self):
         self.revoke()
         self.getData()
 
+    def getAuthCacheFile(self):
+        oldmask = os.umask(0o077)
+        path = Path('/tmp/pcw/{}/{}'.format(self.__class__.__name__, self.namespace))
+        path.mkdir(mode=0o700, parents=True, exist_ok=True)
+        f = path.joinpath('auth.json')
+        f.touch(mode=0o600, exist_ok=True)
+        oldmask = os.umask(oldmask)
```
cfconrad:
```suggestion
        os.umask(oldmask)
```

asmorodskyi: yeah I was about to add this :)

webui/ocw/lib/vault.py
asmorodskyi: ah I see

webui/ocw/lib/vault.py
```diff
@@ -75,7 +81,12 @@ def getCredentials(self):
 
     def getData(self, name=None):
         if self.auth_json is None:
-            self.auth_json = self.getCredentials()
+            if self.use_file_cache:
+                self.auth_json = self.loadFileCache()
+            if self.auth_json is None:
+                self.auth_json = self.getCredentials()
+                if self.use_file_cache:
```
asmorodskyi: something wrong here , before I will start explain you what tricky situations we could have here can you please double check that you haven't mess up with indentation here ?

webui/ocw/lib/vault.py
cfconrad: I will keep `f` it's a 2 line with statement and `f` is quite common here

webui/ocw/lib/vault.py
```diff
@@ -75,7 +81,12 @@ def getCredentials(self):
 
     def getData(self, name=None):
         if self.auth_json is None:
-            self.auth_json = self.getCredentials()
+            if self.use_file_cache:
+                self.auth_json = self.loadFileCache()
```
asmorodskyi: two things possible - or you name your patch wrong or you should call `self.auth_json = self.getCredentials()` unconditional to `use_file_cache`  . because what I see here is not "caching the response from server" , but replacing it

cfconrad: It's only set if it isn't set previously...

webui/ocw/lib/vault.py
cfconrad: don't know what you mean. The code changed slightly now here, but this `if` still exists.

cfconrad:
I will squash later on, but now you could see the differents between the commits..

asmorodskyi:
generally LGTM , just remove unneeded variable assignment to oldmask and I will merge :)

cfconrad:
thx for the hint, done.

PR #43, opened 2019-11-08
Cleanup images on EC2, GCE and Azure

cfconrad:
**Cleanup images on Azure CSP**

Find and delete outdated images on Azure.
Add new configuration section `[cleanup]` this can contain:
```
max-images-per-flavor
max-images-age-hours
azure-storage-account-name
azure-storage-resourcegroup
```

The later two can be specified in a namespace specific subgroup named
`[cleanup.namespace.XXX]`.
This is needed as the storage-account is a worldwide unique name.

 **Cleanup images on Google CSP**

We use the same configuration behavior as for Azure!
`cleanup/max-images-per-flavor` and `cleanup/max-images-age-hours`
which can be namespace configured by changing the section from 'cleanup'
to `cleanup.namespace.XXX`

For common code between different providers I introduced super class
`ocw.lib.provider.Provider`.

 **Cleanup images on AWS CSP**

Use same configuration options as on Google and Azure.


README.md
```diff
@@ -62,6 +62,18 @@ age-hours = NUMBER_OF_HOURS_TO_COUNT_AS_LEFT_OVER
 # Optional section to set a specific receiver of left over notifications for
 # a defined vault namespace. XXX should be replaced with the vault namespace
 to = RECEIPE_ADDRESS_NS_1[, RECEIPE_ADDRESS_NS_2]
+
+
+[cleanup]
+# Specifiy how many images per flavor get keept
```
asmorodskyi:
```suggestion
# Specify how many images per flavor get kept
```

webui/ocw/lib/EC2.py
```diff
@@ -1,23 +1,30 @@
...
+from datetime import datetime
+from datetime import timedelta
+from datetime import timezone
 import boto3
+import re
 import time
-from .vault import EC2Credential
 import logging
 
+logger = logging.getLogger(__name__)
```
asmorodskyi: why ?

webui/ocw/lib/EC2.py
```diff
@@ -1,23 +1,30 @@
...
     __ec2_resource = dict()
     __ec2_client = dict()
     __credentials = None
-    __logging = None
 
     def __new__(cls, vault_namespace):
         if vault_namespace not in EC2.__instances:
             EC2.__instances[vault_namespace] = object.__new__(cls)
             EC2.__instances[vault_namespace].__credentials = EC2Credential(vault_namespace)
-            EC2.__instances[vault_namespace].__logging = logging.getLogger(__name__)
```
asmorodskyi: maybe better keep it here and add vault_namespace to logger name ? so you can differentiate between log messages from different namespaces ?

webui/ocw/lib/EC2.py
```diff
@@ -67,3 +74,51 @@ def list_regions(self):
 
     def delete_instance(self, instance_id):
         self.ec2_resource().instances.filter(InstanceIds=[instance_id]).terminate()
+
+    def cleanup_all(self):
+        response = self.ec2_client().describe_images(Owners=['self'])
+        images = dict()
```
asmorodskyi:
```suggestion
        distinct_images = dict()
```

webui/ocw/lib/EC2.py
```diff
@@ -67,3 +74,51 @@ def list_regions(self):
 
     def delete_instance(self, instance_id):
         self.ec2_resource().instances.filter(InstanceIds=[instance_id]).terminate()
+
+    def cleanup_all(self):
```
asmorodskyi:
```suggestion
    def cleanup_images(self):
```

webui/ocw/lib/EC2.py
asmorodskyi: ignore this , bad idea

webui/ocw/lib/EC2.py
```diff
@@ -67,3 +74,51 @@ def list_regions(self):
...
+        response = self.ec2_client().describe_images(Owners=['self'])
+        images = dict()
+        for img in response['Images']:
+            # 'CreationDate': '2019-10-22T20:40:45.000Z',
+            # 'ImageId': 'ami-00d30c03d17d3db69',
+            # 'Name': 'openqa-SLES12-SP5-EC2.x86_64-0.9.1-BYOS-Build1.55.raw.xz'
+            logger.debug("[EC2] Found image '{}'".format(img['Name']))
```
asmorodskyi: `[EC2]` - isn't we can get same info from logger name ?

webui/ocw/lib/azure.py
```diff
@@ -71,3 +79,79 @@ def list_resource_groups(self):
...
+        containers = block_blob_service.list_containers()
+        for c in containers:
+            self.__logger.debug('Found container {}'.format(c.name))
+            if (re.match('^bootdiagnostics-', c.name)):
+                self.cleanup_bootdiagnostics_container(block_blob_service, c)
+            if (c.name == 'sle-images'):
+                self.cleanup_sle_images_container(block_blob_service, c)
+
+    def cleanup_bootdiagnostics_container(self, bbsrv, container):
+        timeout = datetime.now(timezone.utc) + timedelta(seconds=60*60*24)
```
asmorodskyi:
```suggestion
        timeout = datetime.now(timezone.utc) + timedelta(minutes=24)
```

webui/ocw/lib/EC2.py
asmorodskyi: bad idea too :)

webui/ocw/lib/azure.py
```diff
@@ -71,3 +79,79 @@ def list_resource_groups(self):
...
+        last_modified = container.properties.last_modified
+        generator = bbsrv.list_blobs(container.name)
```
asmorodskyi: can you explain this variable name ?

webui/ocw/lib/db.py
```diff
@@ -178,6 +178,7 @@ def __update_run():
                 instances = [azure_to_local_instance(i, vault_namespace) for i in instances]
                 logger.info("Got %d resources groups from Azure", len(instances))
                 sync_csp_to_local_db(instances, ProviderChoice.AZURE, vault_namespace)
+                Azure(vault_namespace).cleanup_all()
```
asmorodskyi: I would prefer to keep it in separate function

webui/ocw/lib/azure.py
cfconrad: https://azure-storage.readthedocs.io/ref/azure.storage.blob.baseblobservice.html#azure.storage.blob.baseblobservice.BaseBlobService.list_containers

```
Returns a generator to list the containers...
````

File: webui/ocw/lib/db.py (hunk 9 above)
Comment (cfconrad) [reply webui/ocw/lib/db.py]: I thought about it as well. But then I need to implement that loop over CSP's and namespaces again.

File: webui/ocw/lib/EC2.py (hunk 2 above)
Comment (cfconrad) [reply webui/ocw/lib/EC2.py]: I thought it is to complicated to make the logger "object" specific. 
From my point of view this is the more common way and currently we don't use any specific logger config at all.

File: webui/ocw/lib/EC2.py (hunk 6 above)
Comment (cfconrad) [reply webui/ocw/lib/EC2.py]: Yes, this is how a log look like:
```
2019-11-11 10:22:48,185 ocw.lib.EC2  DEBUG    [EC2] Found image 'openqa-SLES12-SP5-EC2.x86_64-0.9.1-On-Demand-Build1.58.raw.xz'
2019-11-11 10:22:48,189 ocw.lib.EC2  DEBUG    [EC2] Found image 'openqa-SLES12-SP5-EC2.x86_64-0.9.1-BYOS-Build1.58.raw.xz'
2019-11-11 10:22:48,190 ocw.lib.EC2  DEBUG    [EC2] Found image 'openqa-SLES12-SP5-EC2.x86_64-0.9.1-BYOS-Build1.56.raw.xz'
2019-11-11 10:22:48,190 ocw.lib.EC2  DEBUG    [EC2] Found image 'openqa-SLES12-SP5-EC2.x86_64-0.9.1-On-Demand-Build1.56.raw.xz'
2019-11-11 10:22:48,191 ocw.lib.EC2  INFO     [EC2] Delete image 'openqa-SLES12-SP5-EC2.x86_64-0.9.1-On-Demand-Build1.56.raw.xz' (ami:ami-0e0f2XXXX77a06b09)
2019-11-11 10:22:48,288 ocw.lib.EC2  INFO     [EC2] Delete image 'openqa-SLES12-SP5-EC2.x86_64-0.9.1-BYOS-Build1.56.raw.xz' (ami:ami-0b39dXXXXd8fd5ad)
```

File: webui/ocw/lib/db.py (hunk 9 above)
Comment (asmorodskyi) [reply webui/ocw/lib/db.py]: and I think it is right way to go ...

File: webui/ocw/lib/EC2.py (hunk 6 above)
Comment (asmorodskyi) [reply webui/ocw/lib/EC2.py]: ok , let's leave it for now . I think it would be more clear to come up with separate PR for this ...

File: webui/ocw/lib/azure.py (hunk 7 above)
Comment (cfconrad) [reply webui/ocw/lib/azure.py]: ok

File: webui/ocw/lib/EC2.py (hunk 3 above)
Comment (cfconrad) [reply webui/ocw/lib/EC2.py]: To advance for now. And I will keep it the same for all other providers.

File: webui/ocw/lib/db.py (hunk 9 above)
Comment (cfconrad) [reply webui/ocw/lib/db.py]: done

File: webui/ocw/apps.py [hunk 10]
```diff
@@ -1,5 +1,24 @@
 from django.apps import AppConfig
+import os
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 class OcwConfig(AppConfig):
     name = 'ocw'
+    ready_called = False
```
asmorodskyi: is it "ready for call" ? or most probably "already called" ?

asmorodskyi: now I get naming ... most suitable I would say will be `init_cron_called` for variable and `init_cron` for function . WDYT ?

webui/ocw/lib/cron.py
```diff
@@ -0,0 +1,100 @@
+from threading import Thread
+from threading import Lock
+from time import sleep
+from datetime import timedelta, datetime, timezone
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class CronJob:
```
asmorodskyi: I think using something like this https://docs.python.org/3/library/sched.html or this https://apscheduler.readthedocs.io/en/stable/ will allow you to remove some big amount of code here

webui/ocw/apps.py
cfconrad: Look here https://docs.djangoproject.com/en/2.2/ref/applications/#django.apps.AppConfig.ready it means that the `ready()` function was called. 
So maybe `ready_fn_called` ?

asmorodskyi: ah I was not aware that you inherit method name . but from what I have read you don't need this variable at all .  https://docs.djangoproject.com/en/2.2/ref/applications/#django.apps.AppConfig.ready clearly saying that method will be called **only** in initialization phase , right ?

cfconrad: There is a note
```
.... , ready might be called more than once. ....
```

webui/ocw/lib/cron.py
cfconrad: `sched` doesn't seems to have periodical calls, so I will take a look to `apscheduler`. It looks promising, the only thing I'm missing is the information if a Job is currently running or not.
But maybe it's not that important or we just need to make a simple flag in `db.update()`

webui/ocw/apps.py
asmorodskyi: > There is a note
> 
> ```
> .... , ready might be called more than once. ....
> ```

ok I see . let's just stick with name which you initially set :) thanks for clarifications ...

webui/ocw/lib/cron.py
cfconrad: done

webui/ocw/apps.py
```diff
@@ -1,5 +1,45 @@
...
+            return
+        self.ready_called = True
+
+        getScheduler().start()
+
+        try:
+            import ocw.lib.db
+            import ocw.lib.cleanup
+
+            ocw.lib.db.init_cron()
```
asmorodskyi: don't you think that it would be more clear just to directly do 
`getScheduler().add_job(cleanup_run, trigger='interval', minutes=60, id='cleanup_all')` here ?

asmorodskyi: this comment is actually for next line but I would assume idea would be the same for this line too

cfconrad: The thing is that the "id", at least for the db updater, is meaning full. And then I would need to share this "name" in two files.
And the second one, is just because it might be needed in the future...

PR #45, opened 2019-11-12
Add migration file for new field `vault_namespace`

cfconrad:
This file belongs to commit commit 7236a6cc177db0c46088738b7233b02c24cd7191


webui/ocw/migrations/0004_auto_20190905_1552.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 2.1.7 on 2019-09-05 15:52
```
asmorodskyi: do we really need to keep auto generated files in git ? can we just regenerate them when needed ?

cfconrad: In general I would agree. But here it is different, as we run upgrades of already existing databases, which correlate to the current code update. So I think it doesn't harm to keep them.

PR #49, opened 2019-11-19
Set Jobs misfire_grace_time to 120s

cfconrad:
Set the jobs default misfire_grace_time to 120s otherwise we might miss
the cleanup job, because the update job is running each minute. And
currently we do not have concurent threads using the same CSP
credentials.

  * https://apscheduler.readthedocs.io/en/latest/modules/job.html


asmorodskyi:
LGTM , but you need to rebase

cfconrad:
done

PR #51, opened 2019-11-25
Increase info printed on httpGet\Post failures

asmorodskyi:


ocw/lib/vault.py
```diff
@@ -63,7 +63,8 @@ def httpGet(self, path):
             else:
                 return r
         except Exception as e:
-            raise ConnectionError('Connection failed - {}: {}'.format(type(e).__name__, str(e)))
+            raise ConnectionError('URL: {} \n Response: {} \n {}: {}'.format(
+                self.url + path, r.status_code, type(e).__name__, str(e)))
```
cfconrad: I would not use `r` uncondional here. 
From my point of view, we are trying to catch exceptions from `requests.get()` and then `r` isn't defined.

I didn't found the specific thing in python docs, but I gave this a try:

```
def throw_ex():
    raise Exception("FOOOOOO")
    return "bla"


try:
    foo = throw_ex()
except Exception as e:
    print(e)
    print(foo)
```
And if gave me:
```
FOOOOOO
  File "<stdin>", line 10, in <module>
NameError: name 'foo' is not defined
```

asmorodskyi: yeah current approach is wrong I will try to find out something better ...

ocw/lib/vault.py
```diff
@@ -58,22 +58,24 @@ def isClientTokenExpired(self):
...
         try:
             r = requests.post(self.url + path, json=data, headers=headers, verify=self.certificate_dir)
+            logger.debug("HTTP POST URL: {},response : {}".format(self.url + path, r.json()))
             if len(r.content) > 0 and 'errors' in r.json():
                 raise ConnectionError(",".join(r.json()['errors']))
             else:
                 return r
         except Exception as e:
-            raise ConnectionError('Connection failed - {}: {}'.format(type(e).__name__, str(e)))
+            raise ConnectionError('{}: {}'.format(type(e).__name__, str(e)))
```
cfconrad:
```suggestion
            logger.debug("httpGet Failed with:", exc_info=True)
            raise ConnectionError('{}: {}'.format(type(e).__name__, str(e)))
```

ocw/lib/vault.py
```diff
@@ -58,22 +58,24 @@ def isClientTokenExpired(self):
...
         try:
             r = requests.get(self.url + path, headers=self.getClientToken(), verify=self.certificate_dir)
+            logger.debug("HTTP GET URL: {},response : {}".format(self.url + path, r.json()))
             if len(r.content) > 0 and 'errors' in r.json():
                 raise ConnectionError(",".join(r.json()['errors']))
             else:
                 return r
         except Exception as e:
-            raise ConnectionError('Connection failed - {}: {}'.format(type(e).__name__, str(e)))
+            raise ConnectionError('{}: {}'.format(type(e).__name__, str(e)))
```
cfconrad:
```suggestion
            logger.debug("httpGet Failed with:", exc_info=True)
            raise ConnectionError('{}: {}'.format(type(e).__name__, str(e)))
```

asmorodskyi:
```
pcw_1    |   File "/pcw/ocw/lib/db.py", line 184, in update_run
pcw_1    |     for region in cfg.getList(['ec2', 'regions'], EC2(vault_namespace).list_regions()):
pcw_1    |   File "/pcw/ocw/lib/EC2.py", line 28, in __new__
pcw_1    |     EC2.__instances[vault_namespace].check_credentials()
pcw_1    |   File "/pcw/ocw/lib/EC2.py", line 33, in check_credentials
pcw_1    |     self.__credentials.renew()
pcw_1    |   File "/pcw/ocw/lib/vault.py", line 108, in renew
pcw_1    |     self.getData()
pcw_1    |   File "/pcw/ocw/lib/vault.py", line 86, in getData
pcw_1    |     self.auth_json = self.getCredentials()
pcw_1    |   File "/pcw/ocw/lib/vault.py", line 163, in getCredentials
pcw_1    |     return self.httpGet(path).json()
pcw_1    |   File "/pcw/ocw/lib/vault.py", line 66, in httpGet
pcw_1    |     raise ConnectionError('Connection failed - {}: {}'.format(type(e).__name__, str(e)))
pcw_1    | ConnectionError: Connection failed - ConnectionError: Connection failed - ConnectionError: 
```
this is how exception looks like with current master . I was trying to remove duplicate information from exception , you trying to push it back .  
Last line starts from `ConnectionError: `  isn't it obvious that something failed ? in two lines above you can see `File "/pcw/ocw/lib/vault.py", line 66, in httpGet` so why I need to repeat again that we are in httpGet ?

ocw/lib/vault.py
asmorodskyi: same here

ocw/lib/vault.py
cfconrad: To have something to search for. But I don't have hard feelings about that extra message.

And I do not see multiple information, cause we hide the actual exception and raise a ConnectionError. So we never see the backtrace from `requests` or so (this is why I added it to debug). And I would keep this logic, so callee can catch only `ConnectionError`.

asmorodskyi: > To have something to search for. But I don't have hard feelings about that extra message.

So you will still have it with my patch too 

> And I do not see multiple information, cause we hide the actual exception and raise a ConnectionError. 

Once again this message from above is what you get when you run current master .  And I would say that name of our exception which we raise ( `ConnectionError` ) and mention of `httpGet` in stacktrace  drop the need to say it again with `httpGet Failed with:`

cfconrad: I was answering the question in a different order. Let me add the quotes :)

PR #52, opened 2019-11-28
remove nginx from pcw because it is not part of pcw

asmorodskyi:


asmorodskyi:
mmm probably we should merge this only after we finish all with containers . for now I can use this branch for testing ...

cfconrad:
I think it's ok...

PR #54, opened 2019-11-29
Set absolute paths for containers

asmorodskyi:


webui/settings.py
```diff
@@ -131,7 +131,7 @@
 
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/2.1/howto/static-files/
-STATIC_ROOT = os.path.join(BASE_DIR, "static/")
+STATIC_ROOT = os.path.join("/pcw", "static/")
```
cfconrad: hm, I'm not sure if it is a good idea to use `/` root dir. This would avoid using pcw out of the box as user.

asmorodskyi: again I see this a fallback to discussion of having ability to run approach on test environments . During work on this docker approach I find out that best practice is to define something like `TEST_ENV=1`  variable and actually have two set of settings - one for prod and one for test environment . And the same as with gitlab patch my intent was to come up with solution which would be ready to use in prod and I fine with having some tricky logic "outside" before we will implement sane use of `TEST_ENV` . 

> hm, I'm not sure if it is a good idea to use / root dir. This would avoid using pcw out of the box as user.

Disagree with this statement . "This will require root permission on machine where you want to run it" would be the right one . And I have root access to all machines where I run pcw . What about you ?

cfconrad: I don't see a need for this. It just makes thinks more complicated!

webui/settings.py
```diff
@@ -87,7 +87,7 @@
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': ':memory:',
+        'NAME': os.path.join('/db', 'db.sqlite3'),
```
cfconrad:
```suggestion
        'NAME': os.path.join(BASE_DIR, 'db', 'db.sqlite3'),
```
And I think it is worth to create a empty directory `db/` in the root of this git repo.

cfconrad:
wasn't able to push to your branch, as agreed lets follow on with this
https://github.com/cfconrad/pcw/pull/56

PR #56, opened 2019-12-03
Store db file in dedicated folder

cfconrad:
This makes it possible to mount a specific folder withhin a docker
setup.


asmorodskyi:
your approach with gitignore looks strange for me , I would like to follow same pattern here . why not just add one line to root .gitignore ?

cfconrad:
I needed that file to be able to get the db folder created and commited.

PR #60, opened 2019-12-12
Update/sync: be more gentle with sporadic errors.

cfconrad:
Do not send email for each exception. Only if we are not able to update
within 3 tries, send a email for notification.

### DIFF
--- a/ocw/lib/db.py
+++ b/ocw/lib/db.py
@@ -4,6 +4,7 @@ from ..models import StateChoice
 from ..models import ProviderChoice
 from django.db import transaction
 from django.utils import timezone
+import time
 import json
 import dateutil.parser
 from .emailnotify import send_mail, send_leftover_notification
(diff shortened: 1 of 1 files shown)


asmorodskyi:
what about same approach but isolated on every provider ? 
Disadvantage of this approach is that when let's say you have a list of namespaces ['azure','ec2','gce'] and gce is problematic one causing exception you will keep updating 'azure' and 'ec2' over and over again no matter that actually you just did this a minute ago

cfconrad:
Thats true, but I thought it might not harm. Anyway I will change it as it is more professional :)

cfconrad:
done

ocw/lib/db.py
```diff
@@ -166,40 +191,32 @@ def update_run():
...
-                instances = [gce_to_local_instance(i, vault_namespace) for i in instances]
-                logger.info("Got %d instances from GCE", len(instances))
-                sync_csp_to_local_db(instances, ProviderChoice.GCE, vault_namespace)
-
-        except Exception as e:
-            logger.exception("Update failed!")
-            send_mail(type(e).__name__ + ' on Update', traceback.format_exc())
+    for vault_namespace in cfg.getList(['vault', 'namespaces'], ['']):
+        for provider in cfg.getList(['vault.namespace.{}'.format(vault_namespace), 'providers'],
+                                    ['ec2', 'azure', 'gce']):
```
asmorodskyi: hm maybe 
```suggestion
                                    ['']):
```

would be safer approach ?

ocw/lib/db.py
```diff
@@ -166,40 +191,32 @@ def update_run():
...
+            for n in range(max_retries):
+                try:
+                    _update_provider(provider, vault_namespace)
+                except Exception:
+                    logger.exception("Update failed for {} in namespace {}".format(provider, vault_namespace))
+                    email_text.add(traceback.format_exc())
+                    time.sleep(5)
+                else:
+                    break
+            else:
```
asmorodskyi: didn't get where is `if` for this `else`

asmorodskyi: ... or `except`

ocw/lib/db.py
```diff
@@ -166,40 +191,32 @@ def update_run():
...
+                error_occured = True
+                send_mail('Error on update {} in namespace {}'.format(provider, vault_namespace),
+                          "\n{}\n".format('#'*79).join(email_text))
 
     send_leftover_notification()
```
asmorodskyi: do we still want to send this ? I would put it under `not error_occured` too

ocw/lib/db.py
cfconrad: Let's not do it within this PR, as it would need a change in pcw.ini as well

ocw/lib/db.py
cfconrad: it's the for :) lovely python...
https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops

ocw/lib/db.py
cfconrad: It could be that one provider was successful and this provider have new left over. So I would like to still send it.

ocw/lib/db.py
asmorodskyi: ok

ocw/lib/db.py
asmorodskyi: ah I see :)

PR #61, opened 2019-12-13
Introduce first pytests for Vault class

cfconrad:


tests/test_vault.py
```diff
@@ -0,0 +1,185 @@
...
+import base64
+
+
+# Global test data
+namespace = 'test-qa'
+host = 'http://foo.bar'
+leases = list()
+
+
+def random_str(stringLength):
```
asmorodskyi: for me it is reinventing the wheel . there are tons of classes providing such functionality . I personaly prefer  something like this https://docs.python.org/3/library/uuid.html , but in case UUID does not fit your model I am sure we can find something what generating just letter string

tests/test_vault.py
```diff
@@ -0,0 +1,185 @@
...
+    letters = string.ascii_letters
+    return ''.join(random.choice(letters) for i in range(stringLength))
+
+
+class MockResponse:
```
asmorodskyi: maybe make sense to put it in separate file ? I am sure that we will use it not only here

tests/test_vault.py
```diff
@@ -0,0 +1,185 @@
...
+    def __init__(self, json_response={}, status_code=200):
+        self.json_response = json_response
+        self.content = json.dumps(json_response)
+        self.status_code = status_code
+
+    def json(self):
+        return self.json_response
+
+
+def mock_post(url, **kwargs):
```
asmorodskyi: am I missing something here or you calling this function only once currently ?

tests/test_vault.py
```diff
@@ -0,0 +1,185 @@
...
+    if (url.startswith('http://foo.bar/v1/auth/userpass/login/')):
+        username = url[38:]
+        password = kwargs['json']['password']
+        if (username == 'devel' and password == 'sag_ich_nicht'):
+            return MockResponse({'auth': {'client_token': 'TOTAL_SECRET_TOKEN', 'lease_duration': 60*15}})
+        return MockResponse({'errors': ['Unknown user']}, status_code=401)
+
+    if (url.startswith('{}/v1/sys/leases/revoke'.format(host))):
+        lease_id = kwargs['json']['lease_id']
+        try:
```
asmorodskyi: hm try/catch inside Mock object ? we need to discuss this ...

tests/test_vault.py
```diff
@@ -0,0 +1,185 @@
...
+            "lease_duration": 3600,
+            "data": {
+                "key_algorithm": "KEY_ALG_RSA_2048",
+                "key_type": "TYPE_GOOGLE_CREDENTIALS_FILE",
+                "private_key_data": base64.b64encode(keydata.encode('UTF-8')).decode(encoding='UTF-8')
+              }
+            })
+
+
+def test_vault(monkeypatch):
```
asmorodskyi: better to split this mega big test to dozen of one-liners where each one will cover some small piece of functionality

tests/test_vault.py
```diff
@@ -0,0 +1,185 @@
...
+url = {}
+user = devel
+password = sag_ich_nicht
+        """.format(host))
+
+    monkeypatch.setattr(requests, "post", mock_post)
+    monkeypatch.setattr(requests, "get", mock_get)
+
+    v = Vault(namespace)
+    assert v.url == 'http://foo.bar'
```
asmorodskyi: in line 116 we defining url as just hostname , how we getting `http://` in `v.url` ?

tests/test_vault.py
asmorodskyi: yeah I missing something :)

tests/test_vault.py
cfconrad: nice, I will use it.

tests/test_vault.py
cfconrad: It comes from the pcw.ini file. I'm checking the reading of that file here.

tests/test_vault.py
cfconrad: Yeah sure, Lets do it when we need it.

tests/test_vault.py
cfconrad: you could also just check, if this value exists in that array...

PR #62, opened 2019-12-16
Enable codecov.io for this project

cfconrad:


tests/test_vault.py
```diff
@@ -160,6 +158,7 @@ def test_AzureCredential(monkeypatch):
     az.renew()
     assert client_id != az.getData()['client_id']
     az.revoke()
+    az.revoke()
```
asmorodskyi: looks strange , can you explain ?

cfconrad: It's about this line https://github.com/cfconrad/pcw/blob/master/ocw/lib/vault.py#L35

cfconrad: sorry, it was about this https://github.com/cfconrad/pcw/pull/62/files#diff-da02d3e55df0f097f1a488bf8fdd22c3R194

This test is for https://github.com/cfconrad/pcw/blob/master/ocw/lib/vault.py#L38 and the following

PR #64, opened 2020-01-06
Fix vault toke TTL handling

cfconrad:
In case the authentication token expire, _all_ created leased which was
created in that session, are expired as well.
From [1] you can found an explanation:
```
Revocation can happen manually via the API, via the vault revoke cli command,
or automatically by Vault. When a lease is expired, Vault will automatically
revoke that lease. When a token is revoked, Vault will revoke all leases that
were created using that token.
```

So we use [2] to increase our token expire time and also make sure that
we use only one token per Key.
As there is a other concept of "Explicit maximum lifetime" [3]. If I
understood this concept right, there is no way to have an unlimited token.

[1] https://www.vaultproject.io/docs/concepts/lease.html
[2] https://www.vaultproject.io/api/auth/token/index.html#renew-a-token
[3] https://www.vaultproject.io/docs/concepts/tokens.html#explicit-max-ttls

### DIFF
--- a/ocw/lib/vault.py
+++ b/ocw/lib/vault.py
@@ -41,11 +41,14 @@ class Vault:
                 raise e
         finally:
             self.auth_json = None
+            self.client_token = None
+            self.client_token_expire = None
 
     def getClientToken(self):
         if self.isClientTokenExpired():
-            r = self.httpPost('/v1/auth/userpass/login/' + self.user, data={'password': self.password})
-            j = r.json()
+            self.client_token = None
+            self.client_token_expire = None
+            j = self.httpPost('/v1/auth/userpass/login/' + self.user, data={'password': self.password}).json()
             self.client_token = {'X-Vault-Token': j['auth']['client_token']}
             self.client_token_expire = datetime.today() + timedelta(seconds=j['auth']['lease_duration'])
         return self.client_token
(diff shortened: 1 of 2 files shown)


ocw/lib/vault.py
```diff
@@ -41,11 +41,14 @@ def revoke(self):
...
+            self.client_token = None
+            self.client_token_expire = None
 
     def getClientToken(self):
         if self.isClientTokenExpired():
-            r = self.httpPost('/v1/auth/userpass/login/' + self.user, data={'password': self.password})
-            j = r.json()
+            self.client_token = None
+            self.client_token_expire = None
+            j = self.httpPost('/v1/auth/userpass/login/' + self.user, data={'password': self.password}).json()
```
asmorodskyi: currently `httpPost` is called 3 times in our code , one time we not collecting return value at all and in 2 other cases we require JSON after all . What if we will just return JSON from `httpPost`  ?

asmorodskyi:
other than minor comment about return format of httpPost all looks fine. You may merge it after applying my comment or even without it :)

ocw/lib/vault.py
cfconrad: I think I will keep it for now. As I find `.json()` more readable. If changing it, I would like to rename the function as well, maybe like `httpJsonGet()` ?

PR #66, opened 2020-01-07
tests: Cover ec2_to_json function from db.py with tests

asmorodskyi:


asmorodskyi:
@cfconrad  it is ready from my side

asmorodskyi:
since I want to write more tests and this PR is blocking me to continue . Also taking into account that we already had brief overview of changes made here I will dare to merge it without review . @cfconrad in case you have some concerns we can work on them later ...

PR #68, opened 2020-01-09
vault: Common HTTP response handling - prepared for JSON

cfconrad:
We always expect JSON as response to our REST API calls.
So we can make a common error handling for httpGet() and httpPost(),
which return json object or raise an exception on error.

### DIFF
--- a/tests/test_vault.py
+++ b/tests/test_vault.py
@@ -8,7 +8,7 @@ import json
 import base64
 from pathlib import Path, PurePath
 from faker import Faker
-
+from requests.exceptions import HTTPError
 
 # Global test data
 namespace = 'test-qa'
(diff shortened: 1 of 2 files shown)


tests/test_vault.py
```diff
@@ -21,13 +21,21 @@
...
+    def __init__(self, json_response=None, status_code=200, content_type='application/json', url='not_given', text=""):
         self.status_code = status_code
+        if json_response is not None:
+            text = json.dumps(json_response)
+        self.content = text
+        self.text = text
+        self.headers = {'content-type': content_type}
+        self.url = url
 
     def json(self):
```
asmorodskyi: where we using this method ?

asmorodskyi:
generally looks good to me , just have concerns about `MockResponse.json` ( see my comment in code )

tests/test_vault.py
cfconrad: We are using it here: https://github.com/cfconrad/pcw/pull/68/files#diff-7d35fc4e9f2b0c45d5681b1c0bbc34a8R77

asmorodskyi: ah ok I just didn't realize that we overloading this method so I need to check prod code and not test code

PR #69, opened 2020-01-28
Fix: max_images_per_flavor usage, treat as int

cfconrad:
This avoid following error messages:
```
  File "./ocw/lib/cleanup.py", line 23, in cleanup_run
    EC2(vault_namespace).cleanup_all()
  File "./ocw/lib/EC2.py", line 121, in cleanup_all
    if (i < len(img_list) - max_images_per_flavor or img['creation_datetime'] < max_images_age):
TypeError: unsupported operand type(s) for -: 'int' and 'str'
```

### DIFF
--- a/ocw/lib/EC2.py
+++ b/ocw/lib/EC2.py
@@ -113,8 +113,7 @@ class EC2(Provider):
             images[key].sort(key=lambda x: LooseVersion(x['build']))
 
         max_images_per_flavor = self.cfgGet('cleanup', 'max-images-per-flavor')
-        max_images_age = datetime.now(timezone.utc) - timedelta(hours=int(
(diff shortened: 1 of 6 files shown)


asmorodskyi:
LGTM , but just as idea - maybe better to have function which return int ? 
In case you disagree just merge ...

cfconrad:
Yes it should be depend on the type of the default value in https://github.com/cfconrad/pcw/blob/master/ocw/lib/provider.py#L11 
Let me give it a try...

cfconrad:
@asmorodskyi wdyt?

asmorodskyi:
> @asmorodskyi wdyt?

awesome !

PR #70, opened 2020-02-07
Fix cleanup images - new image names

cfconrad:
The naming scheme changed from publiccloud images. This patch adopt to
the latest one.

### DIFF
--- a/ocw/lib/EC2.py
+++ b/ocw/lib/EC2.py
@@ -74,40 +74,77 @@ class EC2(Provider):
     def delete_instance(self, instance_id):
         self.ec2_resource().instances.filter(InstanceIds=[instance_id]).terminate()
 
+    def parse_image_name(self, img_name):
+        regexes = [
+                   # openqa-SLES12-SP5-EC2.x86_64-0.9.1-BYOS-Build1.55.raw.xz
+                   re.compile(r'''^openqa-SLES
+                              (?P<version>\d+(-SP\d+)?)
+                              -(?P<flavor>EC2)
+                              \.
+                              (?P<arch>[^-]+)
+                              -
+                              (?P<kiwi>\d+\.\d+\.\d+)
+                              -
+                              (?P<type>(BYOS|On-Demand))
+                              -Build
+                              (?P<build>\d+\.\d+)
+                              \.raw\.xz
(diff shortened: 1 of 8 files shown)


ocw/lib/EC2.py
```diff
@@ -74,6 +74,73 @@ def list_regions(self):
...
+                        -
+                        (?P<flavor>EC2[-\w]+)
+                        -Build
+                        (?P<build>\d+\.\d+)
+                        \.raw\.xz
+                        ''', re.RegexFlag.X)
+        m = re.match(regex, img_name)
+        if m:
+            return {
+                    'key': '-'.join([m.group('version'), m.group('flavor'), m.group('type') or '', m.group('arch')]),
```
asmorodskyi: we have 'type' only here

asmorodskyi: in regex above `BYOS` is called `flavor`  and here same thing is `type` . this is confusing

ocw/lib/EC2.py
```diff
@@ -82,32 +149,21 @@ def cleanup_all(self):
             # 'ImageId': 'ami-00d30c03d17d3db69',
             # 'Name': 'openqa-SLES12-SP5-EC2.x86_64-0.9.1-BYOS-Build1.55.raw.xz'
```
asmorodskyi: this comment should be removed / updated and better also moved into function

ocw/lib/EC2.py
cfconrad: That is because the "flavor" is splitted in two parts.
One is `EC2-HVM` and the other is `BYOS` or not.

It's more clear for azure images. There we have a Flavor `Azure-Standard`, `Azure-Basic` and each with and without `BYOS`. 

In the past, without `BYOS` was called `On-Demand`.

ocw/lib/EC2.py
cfconrad: What I wil show here is, that we get a image  as a dict and the key `Name` contains the name, `ImageId` the id and `CreationDate` the date in that format...

cfconrad: What do you think with a command like
# @see https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_images

PR #71, opened 2020-02-18
Add automatic deletion of Instances

cfconrad:
If an instance has a tag called openqa_ttl the number is used in seconds
to set the max age. If this max age is reached, we delete the instance
automatically.

A instance in state DELETING still produce a email notification. As it
might happen that a delete_instance() just went wrong somehow on the CSP
site and manual interaction is needed.


cfconrad:
FYI @pdostal @ldevuder

ocw/lib/db.py
```diff
@@ -100,7 +103,8 @@ def ec2_to_local_instance(instance, vault_namespace, region):
         instance_id=instance.instance_id,
         state=StateChoice.ACTIVE,
         region=region,
-        csp_info=json.dumps(csp_info, ensure_ascii=False)
+        csp_info=json.dumps(csp_info, ensure_ascii=False),
+        max_age=timedelta(seconds=int(csp_info['tags'].get('openqa_ttl', 0))),
```
asmorodskyi: why tag name is openqa_ttl and field name is max_age ? why we need to keep it inside csp_info and as a separate field at the same time ?

ocw/lib/db.py
```diff
@@ -222,6 +229,48 @@ def update_run():
...
+
+
+def auto_delete_instances():
+    cfg = ConfigFile()
+    for vault_namespace in cfg.getList(['vault', 'namespaces'], ['']):
+        o = Instance.objects
+        o = o.filter(state=StateChoice.ACTIVE, vault_namespace=vault_namespace,
+                     max_age__gt=timedelta(0), age__gte=F('max_age'))
+        for i in o:
+            if not is_delete_instance_allowed(i):
```
asmorodskyi:
```suggestion
            if is_delete_instance_allowed(i):
```
positive branches always easier to read also you can skip one line :D

ocw/lib/db.py
```diff
@@ -222,6 +229,48 @@ def update_run():
...
+                continue
+            logger.info("[{}][{}] TTL expire for instance {}".format(i.provider, i.vault_namespace, i.instance_id))
+            try:
+                delete_instance(i)
+            except Exception as e:
+                msg = "[{}][{}] Deleting instance ({}) failed".format(i.provider, i.vault_namespace, i.instance_id)
+                logger.exception(msg)
+                send_mail(type(e).__name__ + ' on Cleanup', "{}\n\n{}".format(msg, traceback.format_exc()))
```
asmorodskyi: can produce a lot of spam at some point , what if we collect exceptions and send single email at the end ?

ocw/lib/db.py
cfconrad: `max_age` and `ttl` good hint, lets make it equal. What would you prefer?

csp_info is a JSON data string and it is hart to use in `WHERE` statements via SQL.
Actually, I would like to make all "real, mandatory data" into db fields. 
The next thing we should do is "Tags". WDYT?

ocw/lib/db.py
cfconrad: ok

ocw/lib/db.py
asmorodskyi: `ttl` is more accurate.

asmorodskyi: > csp_info is a JSON data string and it is hart to use in `WHERE` statements via SQL.

yeah sure but my idea was that we could remove it from JSON because we are generating it . But after checking code I see that it might be overkill because we have generic logic which blindly copying all tags ...

pdostal:
From my point of view this is OK.

ocw/lib/db.py
cfconrad: done

ocw/lib/db.py
cfconrad: done

ocw/lib/db.py
cfconrad: done

ldevulder:
I'm not a Python expert at all, but all looks good for me ;-)

ocw/lib/db.py
```diff
@@ -222,6 +229,51 @@ def update_run():
         init_cron()
 
 
+def is_delete_instance_allowed(instance):
+    return 'openqa_created_by' in instance.csp_info
+
+
+def delete_instance(instance):
+    if not is_delete_instance_allowed(instance):
+        raise PermissionError('This instance isn\'t managed by openqa')
```
asmorodskyi: not sure that we really want to raise exception here , maybe skipping is just enough ?
But in case we do let's put csp_info , because we use it to decide to raise this exception

cfconrad: I keep the exception but added csp_info

asmorodskyi:
LGTM - waiting for mister Tortuga :)

PR #72, opened 2020-02-20
Fix: Force empty tags instead of None

cfconrad:
If azure resource group doesn't have tags, we force it to an empty dict
instead of None. This is the behavior for GCE and EC2 so we will have it
for azure as well.


cfconrad:
@asmorodskyi FYI

PR #73, opened 2020-02-28
cleanup: Fix minor errors and move generic part into Provider

asmorodskyi:
@cfconrad  I have issues with understanding logic here . so we need to discuss this before I can continue . see my comments in PR

### DIFF
--- a/ocw/lib/EC2.py
+++ b/ocw/lib/EC2.py
@@ -2,9 +2,6 @@ from .provider import Provider
 from .vault import EC2Credential
 from distutils.version import LooseVersion
 from dateutil.parser import parse
-from datetime import datetime
-from datetime import timedelta
-from datetime import timezone
 import boto3
 import re
 import time
@@ -76,8 +73,8 @@ class EC2(Provider):
 
     def parse_image_name(self, img_name):
         regexes = [
-                   # openqa-SLES12-SP5-EC2.x86_64-0.9.1-BYOS-Build1.55.raw.xz
-                   re.compile(r'''^openqa-SLES
+            # openqa-SLES12-SP5-EC2.x86_64-0.9.1-BYOS-Build1.55.raw.xz
+            re.compile(r'''^openqa-SLES
                               (?P<version>\d+(-SP\d+)?)
                               -(?P<flavor>EC2)
                               \.
@@ -90,10 +87,10 @@ class EC2(Provider):
                               (?P<build>\d+\.\d+)
                               \.raw\.xz
                               ''', re.RegexFlag.X),
-                   # openqa-SLES15-SP2.x86_64-0.9.3-EC2-HVM-Build1.10.raw.xz'
-                   # openqa-SLES15-SP2-BYOS.x86_64-0.9.3-EC2-HVM-Build1.10.raw.xz'
-                   # openqa-SLES15-SP2.aarch64-0.9.3-EC2-HVM-Build1.49.raw.xz'
-                   re.compile(r'''^openqa-SLES
+            # openqa-SLES15-SP2.x86_64-0.9.3-EC2-HVM-Build1.10.raw.xz'
+            # openqa-SLES15-SP2-BYOS.x86_64-0.9.3-EC2-HVM-Build1.10.raw.xz'
+            # openqa-SLES15-SP2.aarch64-0.9.3-EC2-HVM-Build1.49.raw.xz'
+            re.compile(r'''^openqa-SLES
                               (?P<version>\d+(-SP\d+)?)
                               (-(?P<type>[^\.]+))?
                               \.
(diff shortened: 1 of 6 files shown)


ocw/lib/azure.py
```diff
@@ -97,13 +97,12 @@ def cleanup_all(self):
                 self.cleanup_sle_images_container(block_blob_service, c)
 
     def cleanup_bootdiagnostics_container(self, bbsrv, container):
-        timeout = datetime.now(timezone.utc) + timedelta(hours=24)
         last_modified = container.properties.last_modified
         generator = bbsrv.list_blobs(container.name)
         for blob in generator:
             if (last_modified < blob.properties.last_modified):
                 last_modified = blob.properties.last_modified
-        if (timeout > last_modified):
```
asmorodskyi: in my understanding this does not make sense . **now** will be always bigger than last_modified because **was** written before you have read it :) so it suppose to be bigger on few millisecond. Or I missing somehting ?

ocw/lib/azure.py
```diff
@@ -137,13 +136,13 @@ def parse_image_name(self, img_name):
                        -
                        (?P<kiwi>\d+\.\d+\.\d+)
                        -
-                       (?P<flavor>Azure[-\w]*)
+                       (?P<flavor>Azure[-\w]*)?
```
asmorodskyi: We did same change in regex for triggering for HPC , so we need to do the same in cleanup logic , right ? We need to think how to bind this logic together ...

ocw/lib/EC2.py
```diff
@@ -141,13 +141,13 @@ def cleanup_all(self):
...
                     'id': img['ImageId'],
-                    })
+                })
             else:
                 logger.error("[EC2][{}] Unable to parse image name '{}'".format(
                     self.__credentials.namespace, img['Name']))
 
         for key in images:
-            images[key].sort(key=lambda x: LooseVersion(x['build']))
+            images[key].sort(key=lambda x: x['creation_datetime'])
```
asmorodskyi: WDYT , will this solve the problem ?

cfconrad: I don't like it. I would make it explicit with min-age, maxage and keep versions.
The idea of keeping the last image per Flavor was, that if you will run something manual on the latest image. you have it available.

ocw/lib/azure.py
cfconrad: Good catch. But you need to include the - as optional

ocw/lib/azure.py
cfconrad: ah funny, so we actually delete the container only if it is older then 24h.
so this is the minage, or?

ocw/lib/provider.py
```diff
@@ -9,20 +12,34 @@ def __init__(self, namespace):
 
     def cfgGet(self, section, field):
         mapping = {
-                'cleanup/max-images-per-flavor': {'default': 1},
-                'cleanup/max-images-age-hours': {'default': 24 * 31},
-                'cleanup/azure-storage-resourcegroup': {'default': 'openqa-upload'},
-                'cleanup/azure-storage-account-name': {'default': 'openqa'},
-                }
+            'cleanup/max-images-per-flavor': {'default': 1},
+            'cleanup/max-images-age-hours': {'default': 24 * 31},
```
asmorodskyi:
```suggestion
            'cleanup/max-image-age-hours': {'default': 24 * 31},
```

ocw/lib/provider.py
```diff
@@ -9,20 +12,34 @@ def __init__(self, namespace):
...
         e = mapping[key]
         namespace_section = '{}.namespace.{}'.format(section, self.__namespace)
         cfg = ConfigFile()
         return type(e['default'])(cfg.get(
-                [namespace_section, field],
-                cfg.get([section, field], e['default'])))
+            [namespace_section, field],
+            cfg.get([section, field], e['default'])))
+
+    def older_min_age(self, age):
```
asmorodskyi:
```suggestion
    def older_than_min_age(self, age):
```

ocw/lib/azure.py
```diff
@@ -137,13 +133,13 @@ def parse_image_name(self, img_name):
                        -
                        (?P<kiwi>\d+\.\d+\.\d+)
                        -
```
cfconrad:
```suggestion

```

ocw/lib/azure.py
```diff
@@ -137,13 +133,13 @@ def parse_image_name(self, img_name):
...
-                       (?P<flavor>Azure[-\w]*)
+                       (?P<flavor>Azure[-\w]*)?
```
cfconrad:
```suggestion
                       (-(?P<flavor>Azure[-\w]*))?
```

asmorodskyi: nice catch ! thanks . fixed

ocw/lib/azure.py
```diff
@@ -137,13 +133,13 @@ def parse_image_name(self, img_name):
                        -
                        (?P<kiwi>\d+\.\d+\.\d+)
                        -
-                       (?P<flavor>Azure[-\w]*)
+                       (-(?P<flavor>Azure[-\w]*))?
```
cfconrad: Plz add 
```
    assert az.parse_image_name('SLES15-SP2-BYOS.aarch64-0.9.3-Azure-Build2.36.vhdfixed.x') == {
            'key': '15-SP2-Azure-BYOS-aarch64',
            'build': '0.9.3-2.36'
            }
```
To test_azure.py

cfconrad: And maybe these too, to be on the safe side:
```
    assert az.parse_image_name('SLES15-SP2-HPC-BYOS.x86_64-0.9.2-Azure-Build2.38.vhd') == {
            'key': '15-SP2-Azure-HPC-BYOS-x86_64',
            'build': '0.9.2-2.38'
            }
    assert az.parse_image_name('SLES15-SP2-BYOS.aarch64-0.9.3-Azure-Build2.36.vhdfixed.x') == {
            'key': '15-SP2-Azure-BYOS-aarch64',
            'build': '0.9.3-2.36'
            }
```

PR #78, opened 2020-03-10
cleanup: Cover Azure.cleanup_sle_images_container() with tests

asmorodskyi:


ocw/lib/provider.py
```diff
@@ -35,7 +35,9 @@ def needs_to_delete_image(self, order_number, image_date):
         if self.older_than_min_age(image_date):
             max_images_per_flavor = self.cfgGet('cleanup', 'max-images-per-flavor')
             max_image_age = image_date + timedelta(hours=self.cfgGet('cleanup', 'max-image-age-hours'))
-            return order_number > max_images_per_flavor or max_image_age < datetime.now(timezone.utc)
+            # order_number is starting from 0 and max_images_per_flavor is **amount**
+            # so we need to increase order_number by 1 to be compare them
+            return order_number+1 > max_images_per_flavor or max_image_age < datetime.now(timezone.utc)
```
cfconrad: I think the most interesting thing is, that the image list must be ordered DESC and starting at index 0 is common.
```suggestion
            return order_number >= max_images_per_flavor or max_image_age < datetime.now(timezone.utc)
```

asmorodskyi: so you think no need to explanation here ?

tests/test_azure.py
```diff
@@ -35,3 +56,39 @@ def test_parse_image_name(monkeypatch):
...
+    monkeypatch.setattr(Azure, 'check_credentials', lambda *args, **kwargs: True)
+    monkeypatch.setattr(BlockBlobService, 'list_blobs', list_blobs_mock)
+    monkeypatch.setattr(BlockBlobService, 'delete_blob', delete_blob_mock)
+    monkeypatch.setattr(Provider, 'cfgGet', mock_cfgGet)
+    az = Azure('fake')
+    az.cleanup_sle_images_container(BlockBlobService(account_name='openqa', account_key='www'), test_name)
+    deleted = ['SLES15-SP2-Azure-HPC.x86_64-0.9.0-Build1.43.vhd', 'SLES15-SP2-Azure-HPC.x86_64-0.9.1-Build1.3.vhd',
+               'SLES15-SP2-BYOS.x86_64-0.9.3-Azure-Build2.36.vhd', 'SLES15-SP2-BYOS.x86_64-0.9.6-Azure-Build1.3.vhd']
+    for item in deleted:
+        assert item in delete_calls[test_name]
```
cfconrad:
```suggestion
    assert delete_calls[test_name] == deleted
```
The order must fit with the algorithm but you also check with this, that we do not delete to much.

PR #79, opened 2020-03-12
Fix: Delete disk and images from upload resource group

cfconrad:
With this patch, we cleanup the disk and image resources from
a resource group specified with azure-storage-resourcegroup in pcw.ini

The cleanup rules for images still apply here.

### DIFF
--- a/ocw/lib/azure.py
+++ b/ocw/lib/azure.py
@@ -1,4 +1,4 @@
-from .provider import Provider
+from .provider import Provider, Image
 from ..lib.vault import AzureCredential
 from azure.common.credentials import ServicePrincipalCredentials
 from azure.mgmt.resource import ResourceManagementClient
@@ -6,7 +6,6 @@ from azure.mgmt.compute import ComputeManagementClient
 from azure.mgmt.storage import StorageManagementClient
 from azure.storage.blob import BlockBlobService
 from msrest.exceptions import AuthenticationError
-from distutils.version import LooseVersion
 import re
 import time
 import logging
@@ -67,6 +66,14 @@ class Azure(Provider):
                 self.sp_credentials(), self.subscription())
         return self.__resoure_mgmt_client
 
+    def get_storage_key(self):
+        resourcegroup = self.cfgGet('cleanup', 'azure-storage-resourcegroup')
+        storage_account = self.cfgGet('cleanup', 'azure-storage-account-name')
+        storage_client = StorageManagementClient(self.sp_credentials(), self.subscription())
+        storage_keys = storage_client.storage_accounts.list_keys(resourcegroup, storage_account)
+        storage_keys = [v.value for v in storage_keys.keys]
+        return storage_keys[0]
+
     def list_instances(self):
         return [i for i in self.compute_mgmt_client().virtual_machines.list_all()]
 
@@ -76,22 +83,48 @@ class Azure(Provider):
     def delete_resource(self, resource_id):
         return self.resource_mgmt_client().resource_groups.delete(resource_id)
 
+    def list_images_by_resource_group(self, resource_group):
+        return self.list_by_resource_group(resource_group,
+                                           filters="resourceType eq 'Microsoft.Compute/images'")
+
+    def list_disks_by_resource_group(self, resource_group):
+        return self.list_by_resource_group(resource_group,
+                                           filters="resourceType eq 'Microsoft.Compute/disks'")
+
+    def list_by_resource_group(self, resource_group, filters=None):
+        return [item for item in self.resource_mgmt_client().resources.list_by_resource_group(
+                        resource_group, filter=filters)]
+
+    def get_keeping_image_names(self, bbsrv):
+        images = list()
+        for item in bbsrv.list_blobs('sle-images'):
+            m = self.parse_image_name(item.name)
+            if m:
+                images.append(Image(item.name, flavor=m['key'], build=m['build'], date=item.properties.last_modified))
+            else:
+                logger.error("[{}] Unable to parse image name '{}'".format(self.__credentials.namespace, item.name))
+
+        return super().get_keeping_image_names(images)
+
     def cleanup_all(self):
         ''' Cleanup all autodateed data which might created during automated tests.'''
-        resourcegroup = self.cfgGet('cleanup', 'azure-storage-resourcegroup')
-        storage_account = self.cfgGet('cleanup', 'azure-storage-account-name')
-        storage_client = StorageManagementClient(self.sp_credentials(), self.subscription())
-        storage_keys = storage_client.storage_accounts.list_keys(resourcegroup, storage_account)
-        storage_keys = [v.value for v in storage_keys.keys]
-        block_blob_service = BlockBlobService(account_name='openqa', account_key=storage_keys[0])
+        block_blob_service = BlockBlobService(account_name='openqa', account_key=self.get_storage_key())
 
+        self.cleanup_bootdiagnostics(block_blob_service)
+
+        keep_images = self.get_keeping_image_names(block_blob_service)
+        self.cleanup_sle_images_container(block_blob_service, keep_images)
+        self.cleanup_disks_from_rg(keep_images)
+        self.cleanup_images_from_rg(keep_images)
+        for i in keep_images:
+            logger.info("Keep image {} ".format(i))
+
+    def cleanup_bootdiagnostics(self, block_blob_service):
         containers = block_blob_service.list_containers()
         for c in containers:
             logger.debug('Found container {}'.format(c.name))
             if (re.match('^bootdiagnostics-', c.name)):
                 self.cleanup_bootdiagnostics_container(block_blob_service, c)
-            if (c.name == 'sle-images'):
-                self.cleanup_sle_images_container(block_blob_service, c.name)
 
     def cleanup_bootdiagnostics_container(self, bbsrv, container):
         last_modified = container.properties.last_modified
@@ -100,7 +133,7 @@ class Azure(Provider):
             if (last_modified < blob.properties.last_modified):
                 last_modified = blob.properties.last_modified
         if (self.older_than_min_age(last_modified)):
-            logger.info("[Azure] Delete container {}".format(container.name))
+            logger.info("Delete container {}".format(container.name))
             if not bbsrv.delete_container(container.name):
                 logger.error("Failed to delete container {}".format(container.name))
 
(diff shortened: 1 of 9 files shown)


cfconrad:
Forgot about the tests :)

ocw/lib/azure.py
```diff
@@ -85,13 +98,19 @@ def cleanup_all(self):
         storage_keys = [v.value for v in storage_keys.keys]
         block_blob_service = BlockBlobService(account_name='openqa', account_key=storage_keys[0])
 
+        keep_images = self.get_keeping_images()
+        self.cleanup_sle_images_container(block_blob_service, keep_images)
+        self.cleanup_disks_from_rg(keep_images)
+        self.cleanup_images_from_rg(keep_images)
+
+        for i in keep_images:
+            logger.info("[Azure] Keep image {} ".format(i))
```
asmorodskyi:
```suggestion
            logger.info("Keep image {} ".format(i))
```
Can we avoid this tag ? This is how log looks like now , and I clearly see from where message coming without it 
```
[36mpcw      |[0m 2020-03-10 17:27:31,553 ocw.lib.azure DEBUG    Found container bootdiagnostics-openqasus-..
[36mpcw      |[0m 2020-03-10 17:27:31,728 ocw.lib.azure DEBUG    Found container bootdiagnostics-openqasus-...
[36mpcw      |[0m 2020-03-10 17:27:31,897 ocw.lib.azure DEBUG    Found container bootdiagnostics-openqasus-...
[36mpcw      |[0m 2020-03-10 17:27:32,064 ocw.lib.azure DEBUG    Found container bootdiagnostics-openqasus-...
[36mpcw      |[0m 2020-03-10 17:27:32,230 ocw.lib.azure DEBUG    Found container bootdiagnostics-openqasus-...
[36mpcw      |[0m 2020-03-10 17:27:32,397 ocw.lib.azure DEBUG    Found container bootdiagnostics-openqasus-...
[36mpcw      |[0m 2020-03-10 17:27:32,564 ocw.lib.azure DEBUG    Found container bootdiagnostics-openqasus-...
[36mpcw      |[0m 2020-03-10 17:27:32,731 ocw.lib.azure DEBUG    Found container bootdiagnostics-openqasus-..
[36mpcw      |[0m 2020-03-10 17:27:32,898 ocw.lib.azure INFO     [Azure] Delete container bootdiagnostics-...
```

asmorodskyi: also logger has formatting within 
```suggestion
            logger.info("Keep image %d ",i)
```

ocw/lib/azure.py
```diff
@@ -85,13 +98,19 @@ def cleanup_all(self):
...
+
         containers = block_blob_service.list_containers()
```
asmorodskyi: in case this collection needed only for `cleanup_bootdiagnostics_container` maybe we should get it inside this function and looping also inside function ? this would make code flow more clear

ocw/lib/azure.py
```diff
@@ -141,32 +160,45 @@ def parse_image_name(self, img_name):
...
+        container_name = 'sle-images'
+        for img in bbsrv.list_blobs(container_name):
             m = self.parse_image_name(img.name)
-            if (m):
-                key = m['key']
-                if key not in images:
-                    images[key] = list()
-
-                logger.debug('[{}]Image {} is candidate for deletion with build {} '.format(
+            if m:
```
asmorodskyi: but you already pass this collection through `parse_image_name` , so why we need to do it second time ?

asmorodskyi: ah wait , that was different collection ... hm we need to discuss this :)

ocw/lib/azure.py
```diff
@@ -141,32 +160,45 @@ def parse_image_name(self, img_name):
...
-            images[key].sort(key=lambda x: LooseVersion(x['build']), reverse=True)
 
-        for img_list in images.values():
-            for i in range(0, len(img_list)):
-                img = img_list[i]
-                if (self.needs_to_delete_image(i, img['last_modified'])):
-                    logger.info("[Azure] Delete image '{}'".format(img['name']))
-                    bbsrv.delete_blob(container_name, img['name'], snapshot=None)
+                if img.name not in keep_images:
+                    logger.info("[Azure] Delete image '{}'".format(img.name))
```
asmorodskyi: wait you just said one line above that this blob not image , also same concern about `[Azure]` tag and not needed `format` call but I think I won't spam PR with mentioning this for every log call which you doing

ocw/lib/azure.py
```diff
@@ -141,32 +160,45 @@ def parse_image_name(self, img_name):
...
+                    bbsrv.delete_blob(container_name, img.name, snapshot=None)
+
+    def cleanup_images_from_rg(self, keep_images):
+        rg = self.cfgGet('cleanup', 'azure-storage-resourcegroup')
+        rg_filter = "resourceType eq 'Microsoft.Compute/images'"
+
+        for item in self.resource_mgmt_client().resources.list_by_resource_group(rg, filter=rg_filter):
```
asmorodskyi: I would go for a function `list_images_by_resource_group`

ocw/lib/azure.py
cfconrad: Ok.

I still think that the vault_namespace might help some times. So I would like to add a Log Adaper which will add this to all logmessages. But this is something for a different PR.

ocw/lib/azure.py
cfconrad: ack

ocw/lib/azure.py
cfconrad: This avoid of deleting arbitrary other disc/images which might be in the rg as well.

ocw/lib/azure.py
cfconrad: Maybe we sould rename `keep_images` into `keep_image_names`.
The Blob, Disk and Image have all the same name.

ocw/lib/azure.py
cfconrad: yes, we can do so

ocw/lib/azure.py
asmorodskyi: vault_namespace would be [kernel-qa] , right ? If so I totally agree to add this and I already followed same approach here https://github.com/cfconrad/pcw/pull/77 . But [Azure] tag does not make much sense when you already have `ocw.lib.azure` just before it

cfconrad: Ack. 
but lets not talk about it in this PR anymore :)

ocw/lib/provider.py
```diff
@@ -50,3 +52,32 @@ def parse_image_name_helper(self, img_name, regex_s, group_key=['version', 'flav
...
+
+        return keep_images
+
+
+class Image:
+
+    def __init__(self, name, flavor, build, date):
+        self.name = name
+        self.flavor = flavor
+        self.build = build
```
jlausuch: what build is this referred to? Shall we differentiate between PC build and kiwi build?

cfconrad: It is the "Build" extracted from image name. Which is `<kiwi>-<PC Build>`. And we use LooseVersion() for parsing this version number and order. So I think not.

ocw/lib/azure.py
```diff
@@ -76,22 +83,48 @@ def list_resource_groups(self):
     def delete_resource(self, resource_id):
         return self.resource_mgmt_client().resource_groups.delete(resource_id)
 
+    def list_images_by_resource_group(self, resource_group):
+        return self.list_by_resource_group(resource_group,
```
asmorodskyi: thanks ! :D

tests/test_azure.py
```diff
@@ -58,37 +62,173 @@ def test_parse_image_name(monkeypatch):
...
 
 
-def test_cleanup_sle_images_container_too_many(monkeypatch):
-    test_name = 'quantity'
+def test_cleanup_sle_images_container(monkeypatch):
+    class FakeBlockBlobService:
+        deleted_blobs = list()
+
+        def list_blobs(self, container):
+            last_modified = None
```
asmorodskyi: why ?  it is already `last_modified=None`  in constructor

tests/test_azure.py
```diff
@@ -58,37 +62,173 @@ def test_parse_image_name(monkeypatch):
...
-    monkeypatch.setattr(BlockBlobService, 'delete_blob', delete_blob_mock)
+    monkeypatch.setattr(Azure, 'get_storage_key', lambda *args, **kwargs: 'FOOXXX')
+    monkeypatch.setattr(Azure, 'get_keeping_image_names', lambda *args, **kwargs: ['a', 'b'])
+    monkeypatch.setattr(Azure, 'cleanup_sle_images_container', lambda *args, **kwargs: None)
+    monkeypatch.setattr(Azure, 'cleanup_disks_from_rg', lambda *args, **kwargs: None)
+    monkeypatch.setattr(Azure, 'cleanup_images_from_rg', lambda *args, **kwargs: None)
+    monkeypatch.setattr(Azure, 'cleanup_bootdiagnostics', lambda *args, **kwargs: None)
+
+    az = Azure('fake')
+    az.cleanup_all()
```
asmorodskyi: and what you proving here ?

tests/test_azure.py
```diff
@@ -58,37 +62,173 @@ def test_parse_image_name(monkeypatch):
...
+    monkeypatch.setattr(Azure, 'compute_mgmt_client', mock_compute_mgmt_client)
+
+    keep_images = ['SLES15-SP2-Azure-HPC.x86_64-0.9.1-Build1.3.vhd']
+    az = Azure('fake')
+    az.cleanup_disks_from_rg(keep_images)
+    assert deleted_disks == ['SLES15-SP2-Azure-HPC.x86_64-0.9.1-Build1.7.vhd']
+
+
+def test_get_keeping_image_names(monkeypatch):
+    class FakeBlockBlobService:
```
asmorodskyi: you keep re-implementing it in every test , isn't it differs on by last_modified date ?

tests/test_provider.py
```diff
@@ -83,3 +85,45 @@ def test_needs_to_delete_image(monkeypatch):
     assert provider.needs_to_delete_image(too_many_images, datetime.now(timezone.utc)) == False
     assert provider.needs_to_delete_image(too_many_images, older_than_min_age) == True
     assert provider.needs_to_delete_image(not_enough_images, older_than_min_age) == False
+
+
+def test_get_keeping_image_names(monkeypatch):
```
asmorodskyi: can you please split this to several tests ?

ocw/lib/azure.py
cfconrad: yw

tests/test_azure.py
cfconrad: Left over. I think I played with times or so... will clean it up

tests/test_azure.py
cfconrad: Coverage

tests/test_azure.py
cfconrad: Most of the time it's a bit different.

tests/test_provider.py
asmorodskyi: may ignore this , not important

tests/test_azure.py
cfconrad: Added a call counter

PR #83, opened 2020-04-21
Rename Vauld.isValid() to Vault.isExpired()

cfconrad:
Simple rename and minimal code change of that method.

### DIFF
--- a/ocw/lib/EC2.py
+++ b/ocw/lib/EC2.py
@@ -25,7 +25,7 @@ class EC2(Provider):
         return EC2.__instances[vault_namespace]
 
     def check_credentials(self):
-        if self.__credentials.isValid():
+        if self.__credentials.isExpired():
             self.__credentials.renew()
             self.__key = None
             self.__secret = None
--- a/ocw/lib/azure.py
+++ b/ocw/lib/azure.py
@@ -4,7 +4,7 @@ from azure.common.credentials import ServicePrincipalCredentials
 from azure.mgmt.resource import ResourceManagementClient
 from azure.mgmt.compute import ComputeManagementClient
 from azure.mgmt.storage import StorageManagementClient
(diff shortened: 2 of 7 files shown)


cfconrad:
Looks like there are some new packages from azure python lib:
```
5.0.0.zip
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-s279j8i2/azure/setup.py", line 60, in <module>
        raise RuntimeError(message)
    RuntimeError:

    Starting with v5.0.0, the 'azure' meta-package is deprecated and cannot be installed anymore.
    Please install the service specific packages prefixed by `azure` needed for your application.
    
    The complete list of available packages can be found at:
    https://aka.ms/azsdk/python/all
    Here's a non-exhaustive list of common packages:

    -  azure-mgmt-compute (https://pypi.python.org/pypi/azure-mgmt-compute) : Management of Virtual Machines, etc.
    -  azure-mgmt-storage (https://pypi.python.org/pypi/azure-mgmt-storage) : Management of storage accounts.
    -  azure-mgmt-resource (https://pypi.python.org/pypi/azure-mgmt-resource) : Generic package about Azure Resource Management (ARM)
    -  azure-keyvault-secrets (https://pypi.python.org/pypi/azure-keyvault-secrets) : Access to secrets in Key Vault
    -  azure-storage-blob (https://pypi.python.org/pypi/azure-storage-blob) : Access to blobs in storage accounts
   

    A more comprehensive discussion of the rationale for this decision can be found in the following issue:
    https://github.com/Azure/azure-sdk-for-python/issues/10646
```

Installing just all 5 still result in a issue like:
```
ocw/lib/azure.py:7: in <module>
    from azure.storage.blob import BlockBlobService
E   ImportError: cannot import name 'BlockBlobService'
```
need to dig into it...

PR #87, opened 2020-09-16
azure: fix AttributeError caused by use of __new__

asmorodskyi:


asmorodskyi:
@cfconrad  check your mail for  " [Openqa-Cloud-Watch] AttributeError on Cleanup in [kernel-qa] " 

this is the fix :) we already discussed this yesterday

ocw/lib/azure.py
```diff
@@ -9,12 +9,19 @@
...
 import time
 import logging
+from typing import Dict
 
 logger = logging.getLogger(__name__)
 
 
 class Azure(Provider):
-    __instances = dict()
+    __instances : Dict[str, "Azure"] = dict()
```
cfconrad, line 18: Why not 
```suggestion
    __instances : Dict[str, Azure] = dict()
```

cfconrad: or
```suggestion
    __instances : Dict[str, Type[Provider] = dict()
```

cfconrad: I didn't found that `"Azure"` hint with google, thats why I'm asking :)

cfconrad (commented):
Plz fix the pip8 hints.

ocw/lib/azure.py
asmorodskyi: `__instances : Dict[str, Type[Provider]] = dict()` 
this is not what I want to because this will allow also to add other siblings of Azure in this array 

` __instances : Dict[str, Azure] = dict()` 
this is not working and throwing some error . I asked in `python` channel for solution and get suggestion to put into a string . see https://chat.suse.de/channel/python?msg=dv8HPSDr7D5ihwLLg

asmorodskyi:
> Plz fix the pip8 hints.

done

PR #89, opened 2020-09-23
ec2: add list clusters support

asmorodskyi:


ocw/lib/EC2.py
```diff
@@ -59,6 +61,21 @@ def ec2_client(self, region='eu-central-1'):
...
         return self.__ec2_client[region]
 
+    def eks_client(self, region='eu-central-1'):
+        if region not in self.__eks_client:
+            self.__eks_client[region] = boto3.client('eks', aws_access_key_id=self.__key,
+                                                     aws_secret_access_key=self.__secret,
+                                                     region_name=region)
+        return self.__eks_client[region]
+
+    def list_clusters(self):
```
cfconrad, line 71: If you take a look to `list_instances()` we only give the vm's of one region.
I somehow would prefer to have some kind of equal behavior in the methods.
So maybe we add a `_all_regions` to the method name then it is clear to me, that it behaves differently then `list_xxxx()`.
WDYT?

ocw/lib/cleanup.py
```diff
@@ -32,5 +32,23 @@ def cleanup_run():
...
+    cfg = ConfigFile()
+    if cfg.has('clusters'):
+        for vault_namespace in cfg.getList(['cleanup', 'namespaces'], ['']):
+            try:
+                orphaned_clusters = EC2(vault_namespace).list_clusters()
+                logger.info("%d clusters found", len(orphaned_clusters))
+                if len(orphaned_clusters):
+                    clusters_str = ' '.join([str(cluster) for cluster in orphaned_clusters])
+                    logger.debug("Full clusters list - %s", clusters_str)
+                    send_mail("EC2 orphaned clusters found", clusters_str)
```
cfconrad, line 45: I think this mail should only go to the corresponding namespace.

ocw/lib/cleanup.py
```diff
@@ -32,5 +32,23 @@ def cleanup_run():
                 send_mail('{} on Cleanup in [{}]'.format(type(e).__name__, vault_namespace), traceback.format_exc())
 
 
+def list_clusters():
+    cfg = ConfigFile()
+    if cfg.has('clusters'):
+        for vault_namespace in cfg.getList(['cleanup', 'namespaces'], ['']):
+            try:
+                orphaned_clusters = EC2(vault_namespace).list_clusters()
```
cfconrad, line 40: Why it is already `orphaned`? You just list clusters and if a test is already running, it isn't a left over? 
What piece I'm missing, plz give me some light!

ocw/lib/cleanup.py
cfconrad: https://github.com/cfconrad/pcw/blob/8ae75a70b43b00d806500b0bf2b3cc2900603052/ocw/lib/emailnotify.py#L54

ocw/lib/cleanup.py
asmorodskyi: yeah you right :) statement is too bold :) the problem that I don't want to invest more time into this to come up with more smart logic where we will try to do some conclusions at least on timestamp :) 
anyway I am fine with any name which you like more than mine :)

cfconrad: hm, what if I say, I don't like to get emails when ever someone run a test.
I really would say we should make some timestamp comparing here.

PR #92, opened 2020-10-26
ec2: delete old snapshots based on settings from pcw.ini

asmorodskyi:


asmorodskyi:
patch obviously need improvements in unit testing area . but I want to understand that after getting LGTM on code part

b10n1k (commented):
just minor things

ocw/lib/vault.py
```diff
@@ -24,7 +24,11 @@ def __init__(self, vault_namespace):
         self.namespace = vault_namespace
         self.password = cfg.get(['vault', 'password'])
         self.certificate_dir = cfg.get(['vault', 'cert_dir'], '/etc/ssl/certs')
-        self.auth_json = None
+        if ConfigFile().getBoolean(['vault', 'use-file-cache']) and self._getAuthCacheFile().exists():
```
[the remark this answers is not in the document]
asmorodskyi: you are right ! thanks !

ocw/lib/cleanup.py
```diff
@@ -47,6 +47,23 @@ def list_clusters():
                     type(e).__name__, vault_namespace), traceback.format_exc())
 
 
+def delete_snapshots():
+    cfg = ConfigFile()
+    if cfg.has('snapshots'):
+        for vault_namespace in cfg.getList(['snapshots', 'namespaces'], ['']):
+            try:
+                delete_after_days = date.today() - timedelta(days=cfg.getInt(['snapshots', 'delete_after_days'], 30))
```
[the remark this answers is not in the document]
asmorodskyi: first of all no need to apologies in PRs - I always keep saying that review process does not have any requirement on expertise in certain project or even technology . You are more than welcome to ask **all** your question and share **all** of 
 your suggestions . I just can not give you any guaranties that everything will be applied :) 
 
Now regarding your certain suggestion here - first of all I just follow the pattern which already existing in `ConfigFile` class . Which already have same type specific getters for List and Boolean . So I would say even if follow your suggestion I would prefer to do it outside this PR and also change existing getters for Boolean and List. Can not judge how much pythonic this approach is in comparison to one which you suggesting ... to sum up - I can not say that I am against your idea but I would like to see some code . At least small code snippet here in this thread at best separate PR for refactor ConfigFIle class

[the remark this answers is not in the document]
b10n1k: so either
`days=int(cfg.get(['snapshots', 'delete_after_days'], 30))`
or create a decorator (keep in mind that i have not dig to the code to deep, so i provide what i had in mind with n example which i run on my interpeter)
```
In [19]: def decor_int(f): 
    ...:     def wrapper(*args, **kwargs): 
    ...:         args = map(int, args) 
    ...:         return f(*args, **kwargs) 
    ...:     return wrapper 
    ...:                                                                                                                                                                                                                                     

In [20]: @decor_int 
    ...: def foo(a): 
    ...:     return a 
    ...:                                                                                                                                                                                                                                     

In [21]: foo('1')                                                                                                                                                                                                                            
Out[21]: 1
```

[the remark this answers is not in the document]
cfconrad: look at this https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.getint
it is the python default lib...
If python would have something like context related return value, then I would go with something implicit, but here I think it is ok.

I don't understand the decorator idea.
We have a function `get()` which return a string by default.
If I now would do a 
```
@decor_int
def get()
```
this function would always return a `int`, or?

ocw/lib/EC2.py
```diff
@@ -76,6 +77,28 @@ def all_clusters(self):
             [clusters.append(cluster) for cluster in response['clusters']]
         return clusters
 
+    def delete_snapshots_older_than(self, delete_date: datetime.date):
```
cfconrad, line 80: Make less complications in my brain for calling that function :)
```suggestion
    def delete_snapshots_older_than(self, age: datetime.timedelta):
```

ocw/lib/EC2.py
```diff
@@ -76,6 +77,28 @@ def all_clusters(self):
...
+        deleted_snapshots = list()
+        for region in regions:
+            response = self.ec2_client(region).describe_snapshots(OwnerIds=['self'])
+            for snapshot in response['Snapshots']:
+                snapshot_start_date = datetime.date(snapshot['StartTime'])
+                if snapshot_start_date < delete_date:
+                    age = date.today() - snapshot_start_date
+                    logger.info("[{}] Deleting snapshot {} in region {} it was {} days old".format(
+                        self.__credentials.namespace, snapshot['SnapshotId'], region, age))
+                    self.ec2_client(region).delete_snapshot(SnapshotId=snapshot['SnapshotId'])
```
cfconrad, line 91: I guess we need to exception handling here as well

ocw/lib/cleanup.py
cfconrad, line 55: In general I would put this configuration into ´[cleanup]` section.
My proposal would be:
```
[cleanup]
ec2-max-snapshot-age-days:
```


And till now, we tried to keep the configurations of such times in a global and namespace specific view.
So a configuration like
```
[snapshots]
namespaces=qam, kernel-qa
delete_after_days = 30
[snapshots.namespace.qam]
delete_after_days = 60
```
should work and have different settings for `qam` and `kernel-qa`. see https://github.com/cfconrad/pcw/blob/d063300e4ecc5d87e9513aefd787f043da417345/ocw/lib/provider.py#L14

ocw/lib/vault.py
```diff
@@ -24,7 +24,11 @@ def __init__(self, vault_namespace):
         self.namespace = vault_namespace
         self.password = cfg.get(['vault', 'password'])
         self.certificate_dir = cfg.get(['vault', 'cert_dir'], '/etc/ssl/certs')
-        self.auth_json = None
+        if cfg.getBoolean(['vault', 'use-file-cache']) and self._getAuthCacheFile().exists():
+            logger.info('Loading cached credentials')
+            self.auth_json = self.loadAuthCache()
+        else:
+            self.auth_json = None
```
cfconrad, line 31: This doesn't look related to this PR

ocw/lib/cleanup.py
```diff
@@ -47,6 +47,23 @@ def list_clusters():
...
+            except Exception as e:
+                logger.exception("[{}] Cleanup snapshots failed!".format(vault_namespace))
+                send_mail('{} on cleanup snapshots in [{}]'.format(
+                    type(e).__name__, vault_namespace), traceback.format_exc())
+
+
 def init_cron():
     getScheduler().add_job(cleanup_run, trigger='interval', minutes=60, id='cleanup_all')
     getScheduler().add_job(list_clusters, trigger='interval', hours=18, id='list_clusters')
+    getScheduler().add_job(delete_snapshots, trigger='interval', hours=18, id='delete_snapshots')
```
cfconrad, line 69: I'm wonder, why you need that extra scheduler? I think it is just a cleanup task

ocw/lib/vault.py
asmorodskyi: this is why it is a separate **commit** , do you also want to create a separate github PR for it ?

ocw/lib/cleanup.py
asmorodskyi: can I add several functions in same scheduler job ? don't think it would be right to put this two different logics into same function . also this gives me more flexibility on different intervals for different cleaners

ocw/lib/EC2.py
asmorodskyi: well it make complications in mine :) can you also highlight how than lines  https://github.com/cfconrad/pcw/pull/92/files#diff-c5779e6ba06607a1398b8f08e0bbbec9563650ba6688e2d3cc6569eddbc7e89eR87 and https://github.com/cfconrad/pcw/pull/92/files#diff-4b964610fda94cfeb813f5a21ed6975cc2ac100ac1546eb851ced96dc488af9bR55 will look like ?

ocw/lib/cleanup.py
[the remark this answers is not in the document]
b10n1k: regarding the decorator i was thinking to put some logic there so for specific variables it will return int. not for all. but the point is not to use the decorator but it looks strange to have a wrapper for int(). it is more clear and 'pure' if in this particular case you have `days=int(cfg.get(['snapshots', 'delete_after_days'], 30))`.

ocw/lib/EC2.py
asmorodskyi: I would actually keep it like this for now .
1. We should not compare this flow with delete_instance . Important difference that in case of exception here we don't have important step which we will miss ( update local DB )
2. I am against catch all exceptions blindly . without changing a lot of other things only what we can do in this case add message to log file , which can be easily overlooked
3. I would prefer more approach which I took in https://github.com/cfconrad/pcw/pull/93 where I catch some known exception for which we already face a problem and re-throw everything else
4. worst case scenario which we will get currently - snapshots will be not deleted which was not happening for a quite some time anyway 🙂

ocw/lib/cleanup.py
asmorodskyi: I am not sure that I can follow you here . 
For me this 
```
 [cleanup]
 ec2-max-snapshot-age-days:
```

is contradicting with this : 
```
[snapshots]
namespaces=qam, kernel-qa
delete_after_days = 30
[snapshots.namespace.qam]
delete_after_days = 60
```

anyway your comment push me to little bit different idea but which might make you happy . Can you please check latest code ?

ocw/lib/cleanup.py
cfconrad: I see it differently. We have a `XXX.cleanup_all()` for all providers. And that function handle all cleanup jobs for orphan artifacts in each provider. 
These `snapshots` are ec2 related and with this only valid for ec2.

ocw/lib/vault.py
cfconrad: no, sorry didn't realized

b10n1k (approved):
LGTM

ocw/lib/cleanup.py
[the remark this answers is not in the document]
asmorodskyi: > regarding the decorator i was thinking to put some logic there so for specific variables it will return int. not for all. but the point is not to use the decorator but it looks strange to have a wrapper for int(). it is more clear and 'pure' if in this particular case you have `days=int(cfg.get(['snapshots', 'delete_after_days'], 30))`.

For me personally , I find it more cleaner to have dedicated method `getInt` than doing same transformation all over the code where we need int from this getter .

ocw/lib/EC2.py
```diff
@@ -160,3 +178,4 @@ def cleanup_all(self):
         for img in [i for i in images if i.name not in keep_images]:
             logger.info("Delete image '{}' (ami:{})".format(img.name, img.id))
             self.ec2_client().deregister_image(ImageId=img.id, DryRun=False)
+        self.delete_outdated_snapshots()
```
cfconrad, line 181: I would move `image deletion` now in a method:
So cleanup all looks like
```
def cleanup_all(self):
   self.cleanup_all_images()
   self.delete_outdated_snapshots()

File: ocw/lib/EC2.py [hunk 8]
```diff
@@ -76,6 +77,23 @@ def all_clusters(self):
             [clusters.append(cluster) for cluster in response['clusters']]
         return clusters
 
+    def delete_outdated_snapshots(self):
+        cleanup_ec2_max_snapshot_age_days = int(self.cfgGet('cleanup', 'ec2-max-snapshot-age-days'))
```
cfconrad, line 81: `int()` isn't needed. You will always get the same type as your `default` value has.

ocw/lib/EC2.py
```diff
@@ -77,6 +78,40 @@ def all_clusters(self):
             [clusters.append(cluster) for cluster in response['clusters']]
         return clusters
 
+    def needs_to_delete_snapshot(self, snapshot, cleanup_ec2_max_snapshot_age_days):
+        delete_older_than = date.today() - timedelta(days=cleanup_ec2_max_snapshot_age_days)
+        if datetime.date(snapshot['StartTime']) < delete_older_than:
+            regexes = [
+                re.compile(r'''^OpenQA upload image$'''),
+                re.compile(r'''^Created by CreateImage\([\w-]*\) for ami-\w* from vol-\w*$''')
```
cfconrad, line 86: nit suggestion
```suggestion
                re.compile(r'''^Created by CreateImage\([\w-]+\) for ami-\w+ from vol-\w+$''')
```

b10n1k (changes requested):
Still not able to do a proper review on this repo. So just small observations from my side

requirements_test.txt
```diff
@@ -7,3 +7,4 @@ flake8
 pytest
 pytest-cov==2.5.0
 codecov
+botocore
```
b10n1k, line 10: this seems to be a dependency on boto3. Why do you redefine it here?

tests/test_ec2.py
```diff
@@ -73,20 +74,86 @@ def mocked_ec2_client():
...
+    correct_description1 = 'OpenQA upload image'
+    correct_description2 = 'Created by CreateImage(jsdkfhsdkj) for ami-sdjhfksdj from vol-sdjfhksdjh'
+    snapshot_to_delete = {'StartTime': old_enough, 'Description': correct_description1}
+    snapshot_to_delete2 = {'StartTime': old_enough, 'Description': correct_description2}
+    not_old_enough = {'StartTime': datetime.now(), 'Description': correct_description1}
+    wrong_description = {'StartTime': old_enough, 'Description': 'DDDDDDDDD'}
+    assert EC2.needs_to_delete_snapshot(snapshot_to_delete, days_to_delete)
+    assert EC2.needs_to_delete_snapshot(snapshot_to_delete2, days_to_delete)
+    assert not  EC2.needs_to_delete_snapshot(not_old_enough, days_to_delete)
+    assert not  EC2.needs_to_delete_snapshot(wrong_description, days_to_delete)
```
b10n1k, line 106: extra whitespace in the two lines above

requirements_test.txt
asmorodskyi: yeah you right , removing it

tests/test_ec2.py
asmorodskyi: removing

ocw/lib/EC2.py
```diff
@@ -77,6 +78,38 @@ def all_clusters(self):
...
+                m = re.match(regex, snapshot['Description'].strip())
+                if m:
+                    return True
+        return False
+
+    def cleanup_snapshots(self):
+        cleanup_ec2_max_snapshot_age_days = self.cfgGet('cleanup', 'ec2-max-snapshot-age-days')
+        if cleanup_ec2_max_snapshot_age_days > 0:
+            regions = self.all_regions()
+            for region in regions:
```
cfconrad, line 99:
```suggestion
            for region in self.all_regions():
```

ocw/lib/EC2.py
```diff
@@ -77,6 +78,38 @@ def all_clusters(self):
...
+            ]
+            for regex in regexes:
+                m = re.match(regex, snapshot['Description'].strip())
+                if m:
+                    return True
+        return False
+
+    def cleanup_snapshots(self):
+        cleanup_ec2_max_snapshot_age_days = self.cfgGet('cleanup', 'ec2-max-snapshot-age-days')
+        if cleanup_ec2_max_snapshot_age_days > 0:
```
cfconrad, line 97: This function has only one path to succeed. I don't see a good reason why not return on the faulty path and avoid with it deep indention. 
```suggestion
        if cleanup_ec2_max_snapshot_age_days < 1:
            return
```

ocw/lib/EC2.py
```diff
@@ -144,6 +177,10 @@ def parse_image_name(self, img_name):
         return self.parse_image_name_helper(img_name, regexes)
 
     def cleanup_all(self):
+        self.cleanup_images()
+        self.cleanup_snapshots()
```
cfconrad, line 181: Nice!

ocw/lib/EC2.py
cfconrad: As the default is `-1` I would actually treat 0 as delete snapshots immediatly.
So this would be my take then:
```suggestion
        if cleanup_ec2_max_snapshot_age_days < 0:
            return
```

PR #93, opened 2020-10-26
ec2: delete_instance should handle case when instance was deleted outside

asmorodskyi:


ocw/lib/EC2.py
```diff
@@ -85,11 +86,14 @@ def all_regions(self):
         return regions
 
     def delete_instance(self, instance_id):
-        instances_list = list(self.ec2_resource().instances.filter(InstanceIds=[instance_id]))
-        if len(instances_list) > 0:
-            instances_list[0].terminate()
-        else:
-            logger.warning("Instance {} is ACTIVE in local DB but does not exists on EC2".format(instance_id))
+        try:
+            self.ec2_resource().instances.filter(InstanceIds=[instance_id]).terminate()
```
jlausuch, line 90: so, this doesn't return a list?

ocw/lib/EC2.py
```diff
@@ -85,11 +86,14 @@ def all_regions(self):
...
+        except ClientError as ex:
+            if ex.response['Error']['Code'] == 'InvalidInstanceID.NotFound':
+                logger.warning("Instance {} is ACTIVE in local DB but does not exists on EC2".format(instance_id))
```
jlausuch, line 93: I am ok with this warning message, but since this function doesn't care about the DB, the reason could be different, for example you have given a wrong `instance_id` value or similar. I would say `Cannot delete instance ID {}. It does not exist on EC2` or something similar

ocw/lib/EC2.py
```diff
@@ -85,11 +86,14 @@ def all_regions(self):
...
 
     def delete_instance(self, instance_id):
-        instances_list = list(self.ec2_resource().instances.filter(InstanceIds=[instance_id]))
-        if len(instances_list) > 0:
-            instances_list[0].terminate()
-        else:
-            logger.warning("Instance {} is ACTIVE in local DB but does not exists on EC2".format(instance_id))
+        try:
+            self.ec2_resource().instances.filter(InstanceIds=[instance_id]).terminate()
+        except ClientError as ex:
```
jlausuch, line 91:
```suggestion
        except ClientError as e:
```

Always seen `e` and never `ex`, but this is just aesthetics, nothing important.

ocw/lib/EC2.py
asmorodskyi: no it returns something which suppose to be converted the list ( `instanceCollection` ) but it is not the point of this PR . 
Before my assumption was that exception happens when object returned by `filter` trying to call terminate. But it appears that `filter` function itself is main cause . It throwing exception when you try to filter by non-existing entity instead of just return empty collection

ocw/lib/EC2.py
asmorodskyi: yeah I would agree that mostly `e` is used but I somehow prefer `ex` :)

ocw/lib/EC2.py
asmorodskyi: done

cfconrad:
LGTM

But I think we also need a sane error handling in https://github.com/cfconrad/pcw/blob/13c33393882e898089daa6ed678e29033c8f07a6/ocw/lib/db.py#L236
or when calling `delete_instance()` here https://github.com/cfconrad/pcw/blob/13c33393882e898089daa6ed678e29033c8f07a6/ocw/views.py#L48. The other call looks good -- so maybe we should put the exception handling directly into `ocw.lib.db.delete_instance()`

PR #95, opened 2020-10-29
db: not update instance which already deleted

asmorodskyi:


asmorodskyi:
this is only real behavioral change https://github.com/cfconrad/pcw/pull/95/files#diff-9ffc01982520c6f142e271f99ca04347ad6ce48cd9b89829e2b6b5b181237569R45 evrything else is just eye candies . WDYT @cfconrad ?

cfconrad:
That we keep it DELETED and continuously writing that error message => ok
But the instance could be there and pcw will not clean it up => bad

=> I'm not sure if this situation ever will happen. For me it would only appear, if there is a instance ID overlap, but I guess this could cause much bigger problems in CSP side.

For me, the most important thing is, delete orphan instances. And this could lead to a instance which we might miss.
Or do I miss something?

PR #96, opened 2020-11-02
ec2: loop over all regions during cleanup

asmorodskyi:
Before this patch EC2 was doing cleanup only in single region.
Now we looping over all available regions

### DIFF
--- a/ocw/lib/EC2.py
+++ b/ocw/lib/EC2.py
@@ -13,6 +13,12 @@ logger = logging.getLogger(__name__)
 
 class EC2(Provider):
     __instances = dict()
+    default_region = 'eu-central-1'
+
+    def __init__(self, namespace: str):
+        super().__init__(namespace)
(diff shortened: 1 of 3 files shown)


jlausuch (approved):
LGTM

ocw/lib/EC2.py
```diff
@@ -13,6 +13,12 @@
 
 class EC2(Provider):
     __instances = dict()
+    default_region = 'eu-central-1'
+
+    def __init__(self, namespace: str):
```
cfconrad, line 18: Hm I'm not sure anymore why I didn't used a `__init__()` function here. But I gues it has something to do with that singleton approach.
Why you don't put this also into the `__new__`?

cfconrad (approved):
LGTM
Not sure with that `__init__` but If you checked it twice, fine for me.

PR #98, opened 2020-11-08
cleanup: take providers from cleanup.namespace.

asmorodskyi:
Before this patch we were taking providers from vault.namespace.<namespace_name>.
This patch changing this to logically expected cleanup.namespace.<namespace_name>


asmorodskyi:
Was verbally agreed with @cfconrad

PR #99, opened 2020-11-11
Add dry_run configuration parameter

asmorodskyi:
This will allow us to test new functionality without
fear to break something in real world. When this parameter
is True every API call to all public cloud providers is replaced
with log message

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,22 @@
+import pytest
+from webui.settings import ConfigFile
+import webui
+import tempfile
+import os
+
+
+def set_pcw_ini(filename, add=''):
+    with open(filename, "w") as f:
+        f.write(add)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def pcw_file():
+    tmpFile = tempfile.mkstemp()
+    webui.settings.CONFIG_FILE = tmpFile[1]
+    set_pcw_ini(tmpFile[1])
+    yield tmpFile[1]
+    os.remove(tmpFile[1])
+
+
+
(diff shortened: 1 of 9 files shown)


pdostal:
This looks very good!

tests/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+    with open(filename, "w") as f:
+        f.write(add)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def generalSetup():
+    tmpFile = tempfile.mkstemp()
+    webui.settings.CONFIG_FILE = tmpFile[1]
+    set_pcw_ini(tmpFile[1])
+    yield tmpFile[1]
```
cfconrad, line 17: not sure about that yield here, can you explain why not `return` or yield in a loop?

tests/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+import tempfile
+
+
+def set_pcw_ini(filename, add=''):
+    with open(filename, "w") as f:
+        f.write(add)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def generalSetup():
```
asmorodskyi, line 13:
```suggestion
def pcw_ini():
```

tests/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+    tmpFile = tempfile.mkstemp()
+    webui.settings.CONFIG_FILE = tmpFile[1]
+    set_pcw_ini(tmpFile[1])
+    yield tmpFile[1]
+    # to make sure that we have clean file before next test
+    set_pcw_ini(tmpFile[1])
```
asmorodskyi, line 19: we don't need this

asmorodskyi: but we need to cleanup a file

cfconrad (approved):
LGTM

PR #102, opened 2020-11-17
configfile: move related logic into ConfigFile

asmorodskyi:


ocw/lib/cleanup.py
```diff
@@ -12,31 +12,30 @@
...
-                send_mail('{} on Cleanup in [{}]'.format(type(e).__name__, vault_namespace), traceback.format_exc())
+                logger.exception("[{}] Cleanup failed!".format(namespace))
+                send_mail('{} on Cleanup in [{}]'.format(type(e).__name__, namespace), traceback.format_exc())
 
 
 def list_clusters():
-    cfg = ConfigFile()
-    if cfg.has('clusters'):
-        for vault_namespace in cfg.getList(['clusters', 'namespaces'], ['']):
+    feature = 'clusters'
```
cfconrad, line 36: I don't see the reason to use a variable here. I think it makes it even less readable.

ocw/lib/emailnotify.py
```diff
@@ -33,28 +33,27 @@ def draw_instance_table(objects):
...
+    subject = ConfigFile().get('notify/subject', 'CSP left overs')
     body_prefix = "Message from {url}\n\n".format(url=build_absolute_uri())
     send_mail(subject, body_prefix + draw_instance_table(o))
 
     # Handle namespaces
     namespaces = list(dict.fromkeys([i.vault_namespace for i in o]))
     for namespace in namespaces:
-        cfg_path = ['notify.namespace.{}'.format(namespace), 'to']
-        if not cfg.has(cfg_path):
+        config_path = 'notify.namespace.{}/to'.format(namespace)
```
cfconrad, line 53: Isn't this also a place for `PCWConfig.get_namespaces_for('notify')`?

cfconrad: hm ok, no it isn't. I thought the get_namespaces is something different.

webui/settings.py
```diff
@@ -165,62 +165,61 @@
...
         try:
-            self.get(name)
+            self.get(config_path)
             return True
         except Exception:
             pass
         return False
 
 
+class PCWConfig():
```
cfconrad, line 207: I think a method which actualy does this namespace lookup would be nice.

A wrapper for this:
```
type(default)(ConfigFile().get('{}/{}'.format(namespace_section, field),
        return type(e['default'])(cfg.get(
                                              ConfigFile().get('{}/{}'.format(section, field), default)))
```
So we have the same logic for all values where a namespace is involved.

ocw/lib/cleanup.py
asmorodskyi: ok

webui/settings.py
asmorodskyi: sorry didn't get you here . you pointing your comment to classname so I am not sure what method do you mean .

webui/settings.py
```diff
@@ -165,62 +165,61 @@
...
+
+    @staticmethod
+    def get_namespaces_for(feature: str, fallback_to_default=False):
```
asmorodskyi, line 210:
```suggestion
    def get_namespaces_for(feature: str):
```

webui/settings.py
cfconrad:
```
[cleanup]
max-images-per-flavor = 1
[cleanup.namespace.XXX]
max-images-per-flavor = 3
```

cfconrad: `PCWcfg.get(namespace, 'cleanup/max-images-per-flavor');`

asmorodskyi: I still see small contradiction here . from one side we speaking about aligning everything to logic which would look for value in this places and in this order : 
1. `{feature}.namespace.{namespace}/property`  ( e.g. `cleanup.namespace.kernel-qa/max-images-per-flavor` )
2. `{feature}/property` ( e.g. `cleanup/max-images-per-flavor` ) 
3. `default/property` ( e.g. `default/max-images-per-flavor` ) 

On other side we have current implementation of `cfgGet` in Provider class where instead of step 3 we looking into special dictionary within this method . and I don't know how would be better to merge this to approaches ? and do we want actually to require to have **any** value defined for step 3 or in some cases we can be fine to not get any value at all ?

cfconrad: Hm, I'm not sure about `3.`. I actually didn't had this on my mind. What if you have a name like `max-age` and then two features with this value... both getting the default from `default/max-age` ?
Can you point me to the case where we want this `default/*` ?

And yes, I would like to have **any** value with a specified default value, even if the default is `None`.

ocw/lib/emailnotify.py
```diff
@@ -65,26 +64,24 @@ def send_leftover_notification():
...
 def send_mail(subject, message, receiver_email=None):
-    cfg = ConfigFile()
-    if not cfg.has('notify'):
+    if not ConfigFile().has('notify'):
         return
 
-    smtp_server = cfg.get(['notify', 'smtp'])
-    port = cfg.get(['notify', 'smtp-port'], 25)
-    sender_email = cfg.get(['notify', 'from'])
+    smtp_server = ConfigFile().get('notify/smtp')
```
cfconrad, line 80:
```suggestion
    smtp_server = PCWConfig.get_feature_property('notify/smtp')
```
All these `get()` should be moved to `get_feature_property()`.

ocw/lib/emailnotify.py
```diff
@@ -33,28 +33,23 @@ def draw_instance_table(objects):
...
     send_mail(subject, body_prefix + draw_instance_table(o))
 
     # Handle namespaces
-    namespaces = list(dict.fromkeys([i.vault_namespace for i in o]))
-    for namespace in namespaces:
-        cfg_path = ['notify.namespace.{}'.format(namespace), 'to']
-        if not cfg.has(cfg_path):
-            continue
-        receiver_email = cfg.get(cfg_path)
+    for namespace in PCWConfig.get_namespaces_for('notify'):
```
asmorodskyi, line 51: @cfconrad  there is some change in the flow . WDYT ?

ocw/lib/emailnotify.py
```diff
@@ -65,26 +60,24 @@ def send_leftover_notification():
 
 
 def send_cluster_notification(namespace, clusters):
-    cfg = ConfigFile()
-    cfg_path = ['notify.cluster.namespace.{}'.format(namespace), 'to']
-    if not cfg.has('notify') or not cfg.has(cfg_path):
+    if not ConfigFile().has('notify'):
```
asmorodskyi, line 63: @cfconrad  also check this

ocw/lib/emailnotify.py
asmorodskyi: done

ocw/lib/emailnotify.py
cfconrad: The result looks similar to me.
The only thing which is missing is:
```
if not receiver_email:
   continue
```
or?

ocw/lib/emailnotify.py
cfconrad: fine for me. You could also  use:
```
 if not PCWConfig.get_feature_property('notify.cluster', 'to', namespace):
    return
```
wdyt?

ocw/lib/emailnotify.py
asmorodskyi: I would call this code duplication . `send_mail` have handling for `reciever_email=None`

ocw/lib/emailnotify.py
asmorodskyi: this would be change in the logic what we have here . No necessarily wrong but not "just code refactoring" anymore : 
1. currently in case you not define `[notify]` section in the config you will exit here 
2. in your version you would rely on totally different logic mixing several places in config file **and** also content of `default_values`  dictionary within a code 
for me current approach  **here** looks more straightforward , but I am open to discuss this . just wanted to make clear that it is not "same way just looks nicer "

ocw/lib/emailnotify.py
```diff
@@ -33,28 +33,22 @@ def draw_instance_table(objects):
...
 def send_leftover_notification():
-    cfg = ConfigFile()
-    if not cfg.has('notify'):
+    if not ConfigFile().has('notify'):
         return
     o = Instance.objects
-    o = o.filter(active=True,
-                 csp_info__icontains='openqa_created_by',
-                 age__gt=timedelta(hours=int(cfg.get(['notify', 'age-hours'], 12))))
+    o = o.filter(active=True, csp_info__icontains='openqa_created_by',
```
asmorodskyi, line 39: we need to move it inside namespaces loop

cfconrad (approved):
Well done! 
LGTM
didn't looked to tests.

PR #104, opened 2020-12-21
pcwconfig: before fallback to default lookup in feature section

asmorodskyi:
This was overlooked during refactoring.
Initial idea was to have such order :
1. <feature>.namespace.<namespace>/providers
2. <feature>/providers
3. [ec2,gce,azure]
Due to mistake second option was not included
in original refactoring PR. This tiny patch
fixing that.

### DIFF
--- a/webui/settings.py
+++ b/webui/settings.py
@@ -247,7 +247,7 @@ class PCWConfig():
     @staticmethod
     def get_providers_for(feature: str, namespace: str):
         return ConfigFile().getList('{}.namespace.{}/providers'.format(feature, namespace),
-                                    ['ec2', 'azure', 'gce'])
+                                    ConfigFile().getList('{}/providers'.format(feature), ['ec2', 'azure', 'gce']))
 
     @staticmethod
     def has(setting: str) -> bool:


webui/settings.py
```diff
@@ -247,7 +247,7 @@ def get_namespaces_for(feature: str) -> list:
     @staticmethod
     def get_providers_for(feature: str, namespace: str):
         return ConfigFile().getList('{}.namespace.{}/providers'.format(feature, namespace),
-                                    ['ec2', 'azure', 'gce'])
+                                    ConfigFile().getList('{}/providers'.format(feature), ['ec2', 'azure', 'gce']))
```
b10n1k, line 250: This seems like it should be just `'{}/providers'.format(feature)` as parameter inside the `ConfigFile().getList` but you call it again the `getList`. i havent checked how the implementation is for `getList` but if this is above is indented maybe the getList could be refactored.

b10n1k (changes requested):
def getList(self, config_path: str, default: list = []) -> list:
 `getList` looks it takes two args here. do i miss something?

webui/settings.py
asmorodskyi: yeah looks like you not getting the purpose . please read commit message of this PR and implementation of getList . in case you will still concerns after this I can try to explain ...

grisu48 (approved):
LGTM!

webui/settings.py
b10n1k: i think i have. but to be honest i found it confusing. it is kinda of if...elif..else but kinda hidden. Anyway i am taking back the request for change but i have one last observation

if i get it right this line                     
`raise LookupError('Missing attribute {} in file {}'.format(config_path, self.filename))` 
is never evaluated. is that ok? Do we care if this list will not found in the config?

asmorodskyi: no we don't , in such case we return **all** providers . exception which you mention is for more generic cases

PR #105, opened 2020-12-22
ec2: add cleanup for VPC

asmorodskyi:


ocw/lib/EC2.py
```diff
@@ -216,6 +218,105 @@ def cleanup_all(self):
...
+                    return False
+            self.log_info('{} has no associated instances. Initializing cleanup of it', vpcId)
+            for gw in vpc.internet_gateways.all():
+                if self.dry_run:
+                    self.log_info('{} will be not deleted due to dry_run mode', gw)
+                else:
+                    self.log_info('Deleting {}', gw)
+                    vpc.detach_internet_gateway(InternetGatewayId=gw.id)
+                    gw.delete()
+            # delete all route table's
```
grisu48, line 240: I'm wondering if it would make sense to refactor those individual routines as functions. Like

    def cleanVpcRouteTable(...)
    def cleanVpcEndoiint(...)

ocw/lib/EC2.py
```diff
@@ -216,6 +218,105 @@ def cleanup_all(self):
...
+                    self.log_info('Deletion of {} skipped due to dry_run mode', subnet)
+                else:
+                    self.log_info('Deleting {}', subnet)
+                    subnet.delete()
+            if self.dry_run:
+                self.log_info('Deletion of VPC skipped due to dry_run mode')
+            else:
+                # finally, delete the vpc
+                self.ec2_resource(region).meta.client.delete_vpc(VpcId=vpcId)
+        except Exception as e:
```
grisu48, line 302: Shouldn't we return `False` in the case a Exception is thrown?

grisu48 (approved):
Minor comments, LGTM!

ocw/lib/EC2.py
asmorodskyi: nice ! thanks for the hint , not sure why I haven't figure out this :)

ocw/lib/EC2.py
asmorodskyi: yeah I think I need to revert this , this happened because I was already returning `True` when no deletion happened due to running instances attached and I needed to distinguish exception thrown case somehow

asmorodskyi: > yeah I think I need to revert this , this happened because I was already returning `True` when no deletion happened due to running instances attached and I needed to distinguish exception thrown case somehow

ok just ignore this :) I got you wrong first time . yeah I will add `return False` in case of exception catch

grisu48: :+1:

ocw/lib/EC2.py
```diff
@@ -216,6 +218,121 @@ def cleanup_all(self):
...
+        return True
+
+    def delete_vpc_subnets(self, vpc):
+        for subnet in vpc.subnets.all():
+            for interface in subnet.network_interfaces.all():
+                if self.dry_run:
+                    self.log_info('Deletion of {} skipped due to dry_run mode', interface)
+                else:
+                    self.log_info('Deleting {}', interface)
+                    interface.delete()
```
okurz, line 259: This seems to be repeated a lot so I suggest to extract a function or method that evaluates the dry-run parameter

tests/test_ec2.py
```diff
@@ -185,9 +186,45 @@ def delete_volume(VolumeId):
...
+
+def test_cleanup_uploader_vpc(monkeypatch):
+    def mocked_ec2_client():
+        pass
+
+    def mocked_get_boolean(config_path, field=None):
+        if config_path == 'default/dry_run':
+            return False
+        else:
+            return True
```
okurz, line 204:
```suggestion
        return config_path != 'default/dry_run'
```

ocw/lib/EC2.py
asmorodskyi: fully agree with you but since this behavior is already used everywhere in project , and this PR is already pretty huge and complex I would prefer to solve this in separate PR . I created an issue to not forget about it https://github.com/SUSE/pcw/issues/107 .

tests/test_ec2.py
asmorodskyi: yeah your version is better , thanks

asmorodskyi:
@okurz

webui/settings.py
```diff
@@ -259,8 +259,16 @@ def has(setting: str) -> bool:
             return False
 
     @staticmethod
-    def getBoolean(config_path: str, default=False) -> bool:
-        value = ConfigFile().get(config_path, default)
+    def getBoolean(config_path: str, namespace: str = None, default=False) -> bool:
```
cfconrad, line 262: I wonder, why we do not change `get_feature_property()` so it handle bool types as well.

And then drop this `getBoolean()`.

cfconrad (commented):
Some nit comments

ocw/lib/EC2.py
```diff
@@ -216,6 +218,121 @@ def cleanup_all(self):
         self.cleanup_images()
         self.cleanup_snapshots()
         self.cleanup_volumes()
+        self.cleanup_uploader_vpcs()
+
+    def delete_vpc(self, region, vpcId):
+        if PCWConfig.getBoolean('cleanup/vpc_only_notify', self._namespace):
```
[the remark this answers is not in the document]
asmorodskyi: agree with separator change . but I find `cleanup/no-vpc` less informative . why you don't like `cleanup/vpc-notify-only` ?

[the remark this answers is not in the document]
cfconrad: I just saw that you also have `cleanup/vpc_cleanup` and then `no-vpc` doesn't say anything...
But I'm not sure about it. as you have some kind of tree in these options. Cause `vpc_only_notify` doesn't work if `vpc_cleanup=false`.

Maybe we can change the code in that manner, that we have as default
```
vpc-cleanup=True   # Find and delete left over VPC's
vpc-notifiy=True      #  Find and notify for left over VPC's
```

and you can set one or the other and get the expected result.

ocw/lib/EC2.py
```diff
@@ -216,6 +218,121 @@ def cleanup_all(self):
...
+        if PCWConfig.getBoolean('cleanup/vpc_cleanup', self._namespace):
+            for region in self.all_regions:
+                response = self.ec2_client(region).describe_vpcs(
+                    Filters=[{'Name': 'isDefault', 'Values': ['false']},
+                             {'Name': 'tag:Name', 'Values': ['uploader-*']}])
+                for vpc in response['Vpcs']:
+                    self.log_info('{} in {} looks like uploader leftover. (OwnerId={}).', vpc['VpcId'], region,
+                                  vpc['OwnerId'])
+                    if not self.delete_vpc(region, vpc['VpcId']) and not self.dry_run:
+                        send_mail('VPC deletion locked by running VMs',
```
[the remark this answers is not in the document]
asmorodskyi: your comment makes me realize that this approach with delete_vpc returns False in several unrelated cases  was bad decision . will redo this part.

webui/settings.py
asmorodskyi: ok

grisu48 (approved):
Still LGTM! :rocket: I like the new functions, which makes it a bit more maintainable IMHO. Good work Anton! :slightly_smiling_face:

webui/settings.py
asmorodskyi: after spending another half-day on this PR I realized that this is enough :) while generally agree with you I would like to move this into separate PR . I create issue to not forget about this https://github.com/SUSE/pcw/issues/108

ocw/lib/EC2.py
```diff
@@ -213,9 +207,133 @@ def parse_image_name(self, img_name):
...
+            self.delete_network_acls(vpc)
+            self.delete_vpc_subnets(vpc)
+            if self.dry_run:
+                self.log_info('Deletion of VPC skipped due to dry_run mode')
+            else:
+                # finally, delete the vpc
+                self.ec2_resource(region).meta.client.delete_vpc(VpcId=vpcId)
+        except Exception as e:
+            self.log_err("{} on VPC deletion. {}", type(e).__name__, traceback.format_exc())
+            send_mail('{} on VPC deletion in [{}]'.format(type(e).__name__, self._namespace), traceback.format_exc())
```
okurz, line 239: This looks like a rather bad style. In general one should not catch "Exception" itself except for very rare cases. And also as the "try" covers the complete body of the method I wonder why you even handle the exception on this level and not on caller level

asmorodskyi: deleting VPC is quite complex thing which includes several step each one can fail in several different way ( network issues , vault/password  problems , various exceptions from AWS API ) idea is that any issues with deleting one VPC should not interrupt process of deleting others . 
I am aware about the fact that catching everything is bad practice , but I don't think it is applicable here . Keep in mind that even if I will remove try/catch here same function will be called again in 1 hour https://github.com/SUSE/pcw/blob/master/ocw/lib/cleanup.py#L45 so from some perspective everything is wrapped in try/catch Exception anyway :).

okurz: well, ok, your decision :)

cfconrad (approved):
LGTM

PR #106, opened 2021-01-18
ec2: add cleanup for volumes

asmorodskyi:


ocw/lib/EC2.py
```diff
@@ -117,6 +117,23 @@ def cleanup_snapshots(self):
...
+            return
+        delete_older_than = date.today() - timedelta(days=cleanup_ec2_max_volumes_age_days)
+        for region in self.all_regions:
+            response = self.ec2_client(region).describe_volumes()
+            for volume in response['Volumes']:
+                if datetime.date(volume['CreateTime']) < delete_older_than:
+                    self.log_info("Deleting volume {} in region {} with CreateTime={}", volume['VolumeId'],
+                                  region, volume['CreateTime'])
+                    if self.dry_run:
+                        self.log_info("Volume deletion of {} skipped due to dry run mode", volume['VolumeId'])
```
b10n1k, line 133: i would assume that even in a dry_run you would like to see the age of the volume.

b10n1k:
@asmorodskyi i was wondering if you have to make a check (to see if the volume to delete is attached to any instane first) before the deletion? but i am not sure as you remove all instance first in `cleanup_all` and maybe it is unnecessary but what happens it would be used alone? just thinking...

pdostal:
> @asmorodskyi i was wondering if you have to make a check (to see if the volume to delete is attached to any instane first) before the deletion? but i am not sure as you remove all instance first in `cleanup_all` and maybe it is unnecessary but what happens it would be used alone? just thinking... 

If I'm not mistaken the API won't allow attached volume to be deleted.

ocw/lib/EC2.py
asmorodskyi: thanks I will update log message

asmorodskyi:
> @asmorodskyi i was wondering if you have to make a check (to see if the volume to delete is attached to any instane first) before the deletion? but i am not sure as you remove all instance first in `cleanup_all` and maybe it is unnecessary but what happens it would be used alone? just thinking...

I am deleting only volumes which are older than 5 days so normally there should be no VMs attached . And yeah in case volumes is in use API will not allow to delete it but on other side your question makes me realize that current approach is too broad and needs to be updated . thanks for your question :)

ocw/lib/EC2.py
asmorodskyi: after double checking - no I won't update . Currently log look like this in dry_run mode : 

```
2021-01-21 20:17:17,218 ocw.lib.EC2  INFO     [qac] Deleting volume vol-0eb80ef28293a9ed7 in region eu-central-1 with CreateTime=2021-01-13 10:08:35.169000+00:00
2021-01-21 20:17:17,218 ocw.lib.EC2  INFO     [qac] Volume deletion of vol-0eb80ef28293a9ed7 skipped due to dry run mode
```
so creation time already printed just one line above . see https://github.com/SUSE/pcw/pull/106/files#diff-c5779e6ba06607a1398b8f08e0bbbec9563650ba6688e2d3cc6569eddbc7e89eR130

asmorodskyi:
> @asmorodskyi i was wondering if you have to make a check (to see if the volume to delete is attached to any instane first) before the deletion? but i am not sure as you remove all instance first in `cleanup_all` and maybe it is unnecessary but what happens it would be used alone? just thinking...

so I introduce support for tag which can protect volume from deletion

asmorodskyi:
ready to be merged from my POV

ocw/lib/EC2.py
```diff
@@ -117,6 +117,32 @@ def cleanup_snapshots(self):
...
+        cleanup_ec2_max_volumes_age_days = PCWConfig.get_feature_property('cleanup', 'ec2-max-volumes-age-days',
+                                                                          self._namespace)
+        if cleanup_ec2_max_volumes_age_days < 0:
+            return
+        delete_older_than = date.today() - timedelta(days=cleanup_ec2_max_volumes_age_days)
+        for region in self.all_regions:
+            response = self.ec2_client(region).describe_volumes()
+            for volume in response['Volumes']:
+                if datetime.date(volume['CreateTime']) < delete_older_than:
+                    self.log_info("Deleting volume {} in region {} with CreateTime={}", volume['VolumeId'],
```
cfconrad, line 130: I would move this line to 135, so we do not log "delete volume" but we don't do it.

ocw/lib/EC2.py
cfconrad: Hm if you like this, then ignore my comment above :)

cfconrad (approved):
Nice work!

PR #109, opened 2021-03-31
Add missing image disk regex

grisu48:
Add missing image regex following [poo#90530](https://progress.opensuse.org/issues/90530)


asmorodskyi:
> currently you have majority of tidy up changes in wrong commit and actual tidy up commit contains single change . please do one of two :
> 
> 1. simply squash commits into one
> 2. move all tidy up changes into correct commit

other than that LGTM

PR #110, opened 2021-03-31
Add GitHub actions

grisu48:
Add GitHub actions and remove Travis


.github/workflows/pcw.yaml
```diff
@@ -0,0 +1,26 @@
...
+name: public cloud watcher
+'on': push
+
+jobs:
+  pcw:
+    name: public cloud watcher
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7]
```
asmorodskyi, line 12:
```suggestion
        python-version: [3.6, 3.8]
```

grisu48: :+1:

asmorodskyi:
awesome :+1:  thanks for this never could reach this ...

.github/workflows/pcw.yaml
```diff
@@ -0,0 +1,26 @@
+---
+
+name: public cloud watcher
+'on': push
+
+jobs:
+  pcw:
+    name: public cloud watcher
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
grisu48: Not sure if those are even supported. I'm following the official instructions here: https://docs.github.com/en/actions/guides/building-and-testing-python

But once it runs we can try it out ;-)

.github/workflows/pcw.yaml
```diff
@@ -0,0 +1,26 @@
+---
+
+name: public cloud watcher
+'on': push
```
ggkioulis, line 4: should this be `'on'` instead of `on`?

grisu48: Both is supported, just with `'on'` yamllint doesn't complain about a boolean statement should be `true` or `false`. But I'll have another check!

grisu48: This one:

    .github/workflows/pcw.yaml
      4:1       warning  truthy value should be one of [false, true]  (truthy)

grisu48: Still doesn't want to :disappointed:

.github/workflows/pcw.yaml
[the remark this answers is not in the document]
asmorodskyi: ok let's follow the flow , we can improve this later ...

asmorodskyi (approved):
LGTM , thanks a lot :+1:

grisu48:
Don't know why the Actions are not scheduled in this PR, but they are scheduled in the [branch](https://github.com/grisu48/pcw/actions/runs/704860628)

PR #118, opened 2021-06-18
ec2: mute volume-in-use errors because they are expected

asmorodskyi:


ocw/lib/EC2.py
```diff
@@ -118,7 +118,6 @@ def cleanup_snapshots(self, cleanup_ec2_max_snapshot_age_days):
 
     def cleanup_volumes(self, cleanup_ec2_max_volumes_age_days):
         delete_older_than = date.today() - timedelta(days=cleanup_ec2_max_volumes_age_days)
-        self.log_dbg("Will delete volumes older than {}", delete_older_than)
```
pdostal, line 121: Why did you remove also this line?

asmorodskyi: This message was added recently when I was investigating why PCW not doing volume cleanup at all. Now I know the reasons ( I was simply using codebase which does not have volume cleanup ) so without this purpose I would treat this message producing too much noise. You will see it on every cleanup loop execution no matter if there is some cleanup needed or not. And I know that this is very bad way of dealing with log messages and that I suppose to simply mute it with proper log level  setting, but scary part that I don't know how exactly it is working in python and didn't have time and will to figure out ...

PR #123, opened 2021-09-20
Improve html look 

cfconrad:
1) Use full width of window for instance table
2) Display email notified as  icon
3) Show csp_info pretty formatted
4) Make "Edit Filter" more visible


cfconrad:
This is how it looks:
Screenshot from 2021-09-20 17-10-29

PR #125, opened 2021-12-03
Cleanup GCE service Accounts

b10n1k:
Implementing and extending the cleanup_all.

In brief the `parser_service_accoun_list`:
- Make a request to get the accosiated emails with `projects/suse-sle-qa`
- for each email a new request returns the endpoint to the service
- and then for each of the accounts perform deletion

`parser_service_account_list` needs to  filter out the irrellative accounts
`get_vault_name` is decorated to return the accounts where their keys have extended the 24 hours

Signed-off-by: ybonatakis <ybonatakis@suse.com>

### DIFF
--- a/ocw/lib/decorators.py
+++ b/ocw/lib/decorators.py
@@ -1,0 +1,6 @@
+def filterService(*args, **kwargs):
+    def wrapper(func):
+        def filter_func(*args):
+            return [email for email in func(*args) if email.startswith(kwargs['name'])]
+        return filter_func
+    return wrapper
--- a/ocw/lib/gce.py
+++ b/ocw/lib/gce.py
@@ -3,33 +3,57 @@ from .provider import Provider, Image
 import googleapiclient.discovery
 from google.oauth2 import service_account
 from dateutil.parser import parse
+from .decorators import filterService
+from googleapiclient.errors import HttpError
 import re
 
 
 class GCE(Provider):
     __instances = dict()
 
-    def __new__(cls, vault_namespace):
-        if vault_namespace not in GCE.__instances:
-            GCE.__instances[vault_namespace] = self = object.__new__(cls)
-            self.__credentials = GCECredential(vault_namespace)
-            self.__compute_client = None
-            self.__project = None
+    def __new__(cls, vault_namespace, *args, **kwargs):
+        if vault_namespace in GCE.__instances:
+            return GCE.__instances[vault_namespace]
+        GCE.__instances[vault_namespace] = object.__new__(cls)
         return GCE.__instances[vault_namespace]
+
+    def __init__(self, namespace):
+        super().__init__(namespace)
+        self.__credentials = GCECredential(namespace)
+        self.__compute_client = None
+        self.__iam_client = None
+        self.__project = None
+        self.credentials = None
 
     def compute_client(self):
         if self.__credentials.isExpired():
             self.__credentials.renew()
             self.__compute_client = None
+            self.credentials = None
         self.__project = self.__credentials.getPrivateKeyData()["project_id"]
         if self.__compute_client is None:
-            credentials = service_account.Credentials.from_service_account_info(
+            self.credentials = service_account.Credentials.from_service_account_info(
                 self.__credentials.getPrivateKeyData()
             )
             self.__compute_client = googleapiclient.discovery.build(
-                "compute", "v1", credentials=credentials, cache_discovery=False
+                "compute", "v1", credentials=self.credentials, cache_discovery=False
             )
         return self.__compute_client
+
+    def iam_client(self):
+        if self.__credentials.isExpired():
+            self.__credentials.renew()
+            self.__iam_client = None
+            self.credentials = None
+        self.__project = self.__credentials.getPrivateKeyData()["project_id"]
+        if self.__iam_client is None:
+            self.credentials = service_account.Credentials.from_service_account_info(
+                self.__credentials.getPrivateKeyData()
+            )
+            self.__iam_client = googleapiclient.discovery.build(
+                "iam", "v1", credentials=self.credentials, cache_discovery=True
+            )
+        return self.__iam_client
 
     def list_instances(self, zone):
         """ List all instances by zone."""
@@ -207,3 +231,119 @@ class GCE(Provider):
                 if "warnings" in response:
                     for w in response["warnings"]:
                         self.log_warn(w["message"])
+        self.cleanup_serviceaccounts()
+
+    def cleanup_serviceaccounts(self):
+        self.log_info("## GCE Vault key cleaning - {} ##".format(self.__project))
+
+        # Get service account emails
+        service_accounts_filtered_by_email = self.get_service_accounts()
+        # For each email found return the ones with associated old date
+        service_accounts_filtered_by_email_filtered = self.filter_by_times(service_accounts_filtered_by_email)
+        self.delete_service_accounts(service_accounts_filtered_by_email_filtered)
+
+    @filterService(name='vaultopenqa')
+    def get_service_accounts(self):
+        ''' Returns a list of the emails of the Service Accounts
+
+        The request is using the resource name of the project associated with
+        the service accounts. This returns a list of dict items, which are
+        also not vault related.
+        The objects returned in chunks which contain a `nextPageToken`
+        to the next page. The `get_service_accounts` parses each page
+        and returns the full list. Subsequently, the list is filtered to return
+        only emails correlated with vault Service Accounts
+        '''
+        service_accounts_filtered_by_email = []
+        req = self.iam_client().projects().serviceAccounts().list(
+            name='projects/suse-sle-qa')
+
+        while True:
+            resp = req.execute()
+            service_accounts_filtered_by_email += [service_account['email']
+                                                   for service_account in resp.get('accounts', [])]
+
+            req = self.iam_client().projects().serviceAccounts().list_next(
+                previous_request=req, previous_response=resp)
+            if req is None:
+                return tuple(set(service_accounts_filtered_by_email))
+
+    def filter_by_times(self, vault_account_tuple, time_limit=24):
+        '''Returns a list of the emails of the Service Accounts filtered by _time_limit_
+
+        Makes a API request to the ServiceAccount to list all the keys associated with
+        the corresponding email of the service.
+        The respond returns a json in the format
+        _{'keys': [
+           {'name': 'projects/suse-sle-qa/serviceAccounts/{ACCOUNT}/keys/{KEY_HASH} # the full path for a key,
+            'validAfterTime': '2021-11-15T12:13:43Z', # age of the key
+            'validBeforeTime': '2021-12-02T12:13:43Z',
+            'keyAlgorithm': 'KEY_ALG_RSA_2048',
+            'keyOrigin': 'GOOGLE_PROVIDED',
+            'keyType': 'USER_MANAGED'}
+         ]}
+
+        Using `validAfterTime` the function calculate which Service Accounts
+        should be deleted. The actual delete function takes the email in its
+        request, so the `filter_by_times` has to returned them
+        corelated email addresses of the `service_resp['keys']`
+
+        Parameters
+        ----------
+        vault_account_tuple : googleapiclient.discovery.Resource, required
+            The IAM Service Account instance
+        time_limit : int
+            the number of hours where the Service Accounts are keep been
+            ignored from the cleanup job
+        '''
+        from datetime import datetime, timedelta
+        time_limit = datetime.now() - timedelta(hours=time_limit)
+        dt_frm = datetime.fromisoformat
+        filtered_service_accounts = []
+        for email in vault_account_tuple:
+            service_resp = self.iam_client().projects().serviceAccounts().keys().list(
+                name='projects/-/serviceAccounts/{}'.format(email)).execute()
+            filtered_service_accounts += [k['name'] for k in service_resp['keys'] if dt_frm(
+                    k['validAfterTime'][:-1]) < time_limit]
+        return list(filter(lambda e: e in str(filtered_service_accounts), vault_account_tuple))
+
+    def delete_service_accounts(self, service_accounts):
+        ''' Deletes a list of vault keys
+
+        Expects _iam_service_ of type *googleapiclient.discovery.Resource*
+        and _service_accounts_ list of strings.
+
+        A short comment from the documentation. - Deleting a service account
+        key does not revoke short-lived credentials
+        that have been issued based on the service account key.
+
+        Parameters
+        ----------
+        service_accounts : list, required
+            A list of keys to delete. This is retrieved by the _accounts_ of
+            the *get_service_accounts* which is the email field.
+
+        Raises
+        ------
+        TypeError
+            This will raised when the _service_accounts_ will not match the expected pattern in the _name_
+        HttpError
+            Http Respond Errors
+        '''
+        if (len(service_accounts) < 1):
+            self.log_info("Nothing to delete")
+        else:
+            for account_email in list(service_accounts):
+                if self.dry_run:
+                    self.log_warn(
+                        "Deletion of vault Service Account {} skipped due to dry run mode"
+                        .format(account_email))
+                else:
+                    try:
+                        # TODO: if deletion needs key removal first
+                        # iam_service.projects().serviceAccounts().keys().delete(
+                        #     name=service_accounts).execute()
+                        self.iam_client().projects().serviceAccounts().delete(
+                            name='projects/-/serviceAccounts/{}'.format(account_email)).execute()
+                    except (TypeError, HttpError) as err:
+                        self.log_err("Fail to delete Service Account {} \n{}".format(account_email, err))
--- a/tests/test_gce.py
+++ b/tests/test_gce.py
@@ -1,9 +1,17 @@
+import pytest
 from ocw.lib.gce import GCE
 from webui.settings import PCWConfig
 from tests.generators import min_image_age_hours, max_image_age_hours
-from tests.generators import mock_get_feature_property
 from tests import generators
 from datetime import datetime, timezone, timedelta
+
+
+@pytest.fixture
+def gce_provider_patched(monkeypatch):
+    from tests.generators import mock_get_feature_property
+
+    monkeypatch.setattr(PCWConfig, 'get_feature_property', mock_get_feature_property)
+    return monkeypatch
 
 
 def test_parse_image_name(monkeypatch):
@@ -58,7 +66,26 @@ class FakeMockImages:
         return self.responses.pop(0)
 
 
-def test_cleanup_all(monkeypatch):
+class FakeServiceAccounts:
+
+    def __init__(self, responses):
+        self.deleted = list()
+        self.responses = responses
+
+    def list(self, *args, **kwargs):
+        return self.responses.pop(0)
+
+    def list_next(self, *args, **kwargs):
+        return None
+
+    def delete(self, *args, **kwargs):
+        pass
+
+    def keys(self,  *args, **kwargs):
+        return self.responses
+
+
+def test_cleanup_all(gce_provider_patched):
     newer_then_min_age = datetime.now(timezone.utc).isoformat()
     older_then_min_age = (datetime.now(timezone.utc) - timedelta(hours=min_image_age_hours+1)).isoformat()
     older_then_max_age = (datetime.now(timezone.utc) - timedelta(hours=max_image_age_hours+1)).isoformat()
@@ -84,10 +111,11 @@ def test_cleanup_all(monkeypatch):
 
     def mocked_compute_client():
         pass
-    mocked_compute_client.images = lambda *args, **kwargs: fmi
-    monkeypatch.setattr(GCE, 'compute_client', lambda self: mocked_compute_client)
 
-    monkeypatch.setattr(PCWConfig, 'get_feature_property', mock_get_feature_property)
+    mocked_compute_client.images = lambda *args, **kwargs: fmi
+    gce_provider_patched.setattr(GCE, 'compute_client', lambda self: mocked_compute_client)
+    # Skips cleanup_serviceaccounts
+    gce_provider_patched.setattr(GCE, 'cleanup_serviceaccounts', lambda self: True)
 
     gce = GCE('fake')
     generators.max_images_per_flavor = 2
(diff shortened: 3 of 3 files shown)


ocw/lib/gce.py
```diff
@@ -207,3 +209,112 @@ def cleanup_all(self):
                 if "warnings" in response:
                     for w in response["warnings"]:
                         self.log_warn(w["message"])
+        self.cleanup_vault_keys()
+
+    def cleanup_vault_keys(self):
+        self.log_info("## GCE Vault key cleaning - %s ##" % self.__project)
+        credentials = service_account.Credentials.from_service_account_info(
```
asmorodskyi, line 216:
```
credentials = service_account.Credentials.from_service_account_info(
            self.__credentials.getPrivateKeyData()
        )
        service = googleapiclient.discovery.build('iam', 'v1', credentials=credentials)
```

please drop this and just use https://github.com/SUSE/pcw/pull/125/files#diff-0dfc4095c73002d356e289d8edf872dcb8a6624ba9adc082c0c331aa9789bb9cR22 instead . you duplicating the code

ocw/lib/gce.py
```diff
@@ -207,3 +209,112 @@ def cleanup_all(self):
                 if "warnings" in response:
                     for w in response["warnings"]:
                         self.log_warn(w["message"])
+        self.cleanup_vault_keys()
+
+    def cleanup_vault_keys(self):
+        self.log_info("## GCE Vault key cleaning - %s ##" % self.__project)
```
asmorodskyi, line 215: this comment applies to whole PR . please read https://realpython.com/python-string-formatting/ . in project we currently using `#2` . I am considering moving to `#3` some day but for sure against fallback to `#1`

No about this specific message . You are not cleaning any Vault keys here you cleaning service accounts in google cloud . this also refers to function name

ocw/lib/gce.py
```diff
@@ -207,3 +209,112 @@ def cleanup_all(self):
...
+        to the next page. The `parser_service_account_list` parses each page
+        and returns the full list. Subsequently, the list is filtered to return
+        only emails correlated with vault
+
+        Parameters
+        ----------
+        iam_service : googleapiclient.discovery.Resource, required
+            The IAM Service Account instance
+        '''
+        vault_emails = []
```
asmorodskyi, line 265: what is "vault email" ?

ocw/lib/gce.py
```diff
@@ -207,3 +209,112 @@ def cleanup_all(self):
...
+        )
+        service = googleapiclient.discovery.build('iam', 'v1', credentials=credentials)
+        # now get service accounts
+        vault_emails = self.parser_service_account_list(service)
+        # for each vault account find the associated names with old keys
+        vault_keys = self.get_vault_name(service, vault_emails)
+
+        self.delete_vault_keys(service, vault_keys)
+
+    @filterAfterTime(hours=24)
```
asmorodskyi, line 227: can you elaborate benefit of using decorator pattern here ? AFAIK decorator should be used when you need to do same thing *in several classes not linked by inheritance* than you can save yourself from *code duplication* . Can not find anything related to this topic here . In a fact you just break single flow into two pieces and now it is hard to understand what actually going on with data ...

asmorodskyi (changes requested):
1. tests are failing 
2. not correct use of decorator design pattern
3. confusion around functions/variable names all over PR ( we not suppose to do any cleanup in vault but cleanup of service accounts in gce )

asmorodskyi (changes requested):
1. tests are failing 
2. inappropriate use of decorator design pattern 
3. confusion in functions / variable names ( we suppose to cleanup not vault but google cloud service accounts )

asmorodskyi:
sorry for spam :) after posting first comment I got 500 from github so I decided to do it again ...

ocw/lib/gce.py
b10n1k: i did that in the very beginning and it was failing. i didnt overthink and i left it when i see that it was working. if i fix the rest i might take a look. but until then i am not going to change it as i am afraid that it will require additional changes. i have to investigate why it wasnt working and what changes it needs

ocw/lib/gce.py
b10n1k: to be honest, i dont know how to answer this question and i try to avoid them in general (many reason out of the scope here). Everyone should do their research. for instance decorators do not apply _in several classes_ only. But i will tell you that. some small portion of my thinking is that `filterAfterTime` will apply to ec2 and azure later. Unfortunately i dont have idea how it will look like there so i do not provide the one-off solution. i expect that it will require some changes when i will have clear picture from the other providers.

> In a fact you just break single flow into two pieces and now it is hard to understand what actually going on with data ...

Thats not true. in fact this design is very common in many frameworks (django, Spring, etc). the name of the functions is also self-explanatory. The `get_vault_name` requests and returns the `['keys']` and then you want to `filterAfterTime`, to get the _old_ ones.  `get_vault_name` will never have to change. if believe that if you see what `filterAfterTime` does you will notice that import different packages and does quite different operation, which do not belong inside `get_vault_name` as you would expect from such function.

ocw/lib/gce.py
b10n1k: the take out from the `service_account['email']`

ocw/lib/gce.py
asmorodskyi: sure thing it is up to you to decide the order , just saying that we will need to revisit this before merging in any case or find really good reasons to keep this duplication

ocw/lib/gce.py
asmorodskyi: service account is entity from google cloud `Credentials` , right ? how it is related to vault ?

ocw/lib/gce.py
asmorodskyi: > to be honest, i dont know how to answer this question and i try to avoid them in general (many reason out of the scope here)

sorry I didn't get this part 

> some small portion of my thinking is that filterAfterTime will apply to ec2 and azure later.

currently decorators are not used in this project at all . this PR is about adding google service account cleanup . I would prefer that we will concentrate on this without doing several things at the same time . I am fine to add decorators to the project but I can not accept having this pattern in one single random place while whole project not using same approach . Please solve : 
> Unfortunately i dont have idea how it will look like 

and provide " one-off solution" in separate PR.

ocw/lib/decorators.py
```diff
@@ -0,0 +1,20 @@
+def filterAfterTime(*args, **kwargs):
+    def wrapper(func):
+        from datetime import datetime, timedelta
+        time_limit = datetime.now() - timedelta(hours=kwargs['hours'])
+
+        def time_filtered_list(*args, **kwargs):
+            dat = func(*args)
+            dt_frm = datetime.fromisoformat
+            return [k['name'] for k in dat if dt_frm(
+                    k['validAfterTime'][:-1]) < time_limit]
```
asmorodskyi, line 10: I assume `validAfterTime` is field from Google service account field ?

asmorodskyi:
just to sum up : 
- cleanup_vault_keys
-  vault_email
- vault_keys
- delete_vault_keys 
- get_vault_name
- vault_account_tuple
- delete_vault_keys
all this variable/function names have nothing to do with Vault so their names are confusing . 
Just to give you a hint . In `delete_vault_keys`  [you are deleting](https://github.com/SUSE/pcw/pull/125/files#diff-0dfc4095c73002d356e289d8edf872dcb8a6624ba9adc082c0c331aa9789bb9cR316) using call to `googleapiclient.discovery.Resource.projects().serviceAccounts().keys().delete` which clearly saying that you actually deleting google *service account* keys which has zero linkage to Vault

ocw/lib/gce.py
```diff
@@ -207,3 +209,112 @@ def cleanup_all(self):
...
+            for vault_key in list(vault_keys):
+                if self.dry_run:
+                    self.log_warn(
+                        "Deletion of vault key {} skipped due to dry run mode",
+                        vault_key
+                    )
+                else:
+                    try:
+                        iam_service.projects().serviceAccounts().keys().delete(
+                            name=vault_keys).execute()
```
cfconrad, line 317:
```suggestion
                            name=vault_key).execute()
```

ocw/lib/gce.py
```diff
@@ -207,3 +209,112 @@ def cleanup_all(self):
...
+            vault_emails += [service_account['email']
+                             for service_account in resp.get('accounts', [])]
+
+            req = iam_service.projects().serviceAccounts().list_next(
+                previous_request=req, previous_response=resp)
+            if req is None:
+                break
+        return tuple(set(vault_emails))
+
+    def delete_vault_keys(self, iam_service, vault_keys):
```
cfconrad, line 280: I think we can give this a general name, as we just call delete key and nothging special. The special things regarding vault are on the `collect` side.
```suggestion
    def delete_keys(self, iam_service, keys):
```

ocw/lib/gce.py
```diff
@@ -207,3 +209,112 @@ def cleanup_all(self):
...
+            An iterable object with strings of the Email address associated to the service accounts.
+        '''
+        keys = []
+        for email in vault_account_tuple:
+            service_resp = service.projects().serviceAccounts().keys().list(
+                name='projects/-/serviceAccounts/%s' % email).execute()
+            keys += [k for k in service_resp['keys']]
+        return keys
+
+    @filterService(name='vaultopenqa')
```
cfconrad, line 248: Is this decorator configurable by the caller of `paresr_service_account_list()` ?

ocw/lib/decorators.py
```diff
@@ -0,0 +1,20 @@
...
+        return time_filtered_list
+    return wrapper
+
+
+def filterService(*args, **kwargs):
```
cfconrad, line 15: Maybe `filter_colllection_starts_with` and isn't it the same as https://docs.python.org/3/library/functions.html#filter

ocw/lib/decorators.py
cfconrad: yeah for such common function, I think a you would need to introduce a common class object or just do it in `gce.py`

cfconrad (changes requested):
Nice that you introduced some more python code styles, but I think we should write this code as easy as possible.
 1) collect all emails `l = parser_service_account_list()`
 2) filter list for deleteable accounts (regex read from config) `l = deletable_accounts(l)`
 3) collect all keys `k = get_vault_name(l)`
 4) filter old keys (age read from config) `k_old = only_old_keys(k)
 5) delete all keys `delete_vault_keys(k_old);
 6) filter all service accounts older then XXX (age read from config) `old_l = old_accounts(l)`
 7) Delete all old service accounts `delete_service_accounts(old_l)`

1-5 is what you did, just extract the vault specific in functions and I would read the parameters from the config file

6+7 are there, to cleanup the services accounts it self. I didn't saw it, maybe I missed it.

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *changes applied and waiting for feedback*

ocw/lib/decorators.py
b10n1k: *changes applied and waiting for feedback*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

b10n1k:
I have changed names in the last commit. i also replace a function with `filter_email_service_accounts` which also removes one of the decorators. 
I gave up of the _credentials_ issue as for now. @asmorodskyi had a good argument but that might need additional changes overall. A workaround could be to destroy the credentials as soon as the cleanup ends or created it with really short lifetime. 

Now the `delete_vaultopenqa_service_accounts` deletes accounts. However i am struggling to verify the operation. Even manual. 

let me know if i am in the right direction.

ocw/lib/decorators.py
b10n1k: i would say yes but from POV is just another way to choice of implementation. the decorator has different 'scope'. one is a design pattern, the other is python build-in function. So _same_ is relative

ocw/lib/gce.py
b10n1k: it can become obviously, but i am not sure what do you have in mind exactly.

ocw/lib/gce.py
b10n1k: i havent started the writing of the testing yet. i am going to do it as a very last step. unless..? Do you find good reason to start with it now?

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
                 if "warnings" in response:
                     for w in response["warnings"]:
                         self.log_warn(w["message"])
+        self.cleanup_vaultopenqa_serviceaccounts()
+
+    def cleanup_vaultopenqa_serviceaccounts(self):
```
asmorodskyi, line 220:
```suggestion
    def cleanup_serviceaccounts(self):
```

ocw/lib/gce.py
```diff
@@ -3,31 +3,39 @@
...
+
     def compute_client(self):
         if self.__credentials.isExpired():
             self.__credentials.renew()
             self.__compute_client = None
+            self.credentials = None
         self.__project = self.__credentials.getPrivateKeyData()["project_id"]
         if self.__compute_client is None:
-            credentials = service_account.Credentials.from_service_account_info(
+            self.credentials = service_account.Credentials.from_service_account_info(
```
asmorodskyi, line 34:
```
def iam_client(self):
        if self.__credentials.isExpired():
            self.__credentials.renew()
            self.__compute_client = None
        self.__project = self.__credentials.getPrivateKeyData()["project_id"]
        if self.__iam_client is None:
            credentials = service_account.Credentials.from_service_account_info(self.__credentials.getPrivateKeyData()
            )
            self.__iam_client = googleapiclient.discovery.build(
                "iam", "v1", credentials=credentials, cache_discovery=False
            )
        return self.__iam_client
```

this is expected version . Your approach  will lead to some problems which we already had in PCW and which cause a lot issues with stalled credentials and use  already expired credentials. 

BTW `iam_client` and `compute_client` are good candidates for use of your favorite decorator pattern . this can be done as next step

ocw/lib/gce.py
```diff
@@ -3,31 +3,39 @@
...
+            return GCE.__instances[vault_namespace]
+        GCE.__instances[vault_namespace] = object.__new__(cls)
         return GCE.__instances[vault_namespace]
 
+    def __init__(self, namespace):
+        self.__credentials = GCECredential(namespace)
+        self.__compute_client = None
+        self.__project = None
+        self.credentials = None
+        super().__init__(namespace)
```
asmorodskyi, line 25: this is always best practice and also here it is quite necessary AFAIK to call parent's init as very first thing to do

ocw/lib/gce.py
```diff
@@ -3,31 +3,39 @@
...
     __instances = dict()
 
-    def __new__(cls, vault_namespace):
-        if vault_namespace not in GCE.__instances:
-            GCE.__instances[vault_namespace] = self = object.__new__(cls)
-            self.__credentials = GCECredential(vault_namespace)
-            self.__compute_client = None
-            self.__project = None
+    def __new__(cls, vault_namespace, *args, **kwargs):
+        if vault_namespace in GCE.__instances:
```
asmorodskyi, line 15: if you will check `azure` and `ec2` you will see that they also have `__new__` and `__init__` and I can tell you that before they were the same as `GCE` class . Actually I kinda surprised that GCE still use old stile . So generally +1 to this change . But you did it a bit differently than it is done in `azure` / `ec2` . I can not find any issue with THAT approach , but I would like to discuss this with you ...

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
                 if "warnings" in response:
                     for w in response["warnings"]:
                         self.log_warn(w["message"])
+        self.cleanup_vaultopenqa_serviceaccounts()
+
+    def cleanup_vaultopenqa_serviceaccounts(self):
+        self.log_info(f"## GCE Vault key cleaning - {self.__project} ##")
```
asmorodskyi, line 221:
```suggestion
        self.log_info(f"## GCE Service account cleaning - {self.__project} ##")
```

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
...
+        filtered_service_accounts = []
+        for email in vault_account_tuple:
+            service_resp = service.projects().serviceAccounts().keys().list(
+                name='projects/-/serviceAccounts/%s' % email).execute()
+            filtered_service_accounts += [k['name'] for k in service_resp['keys'] if dt_frm(
+                    k['validAfterTime'][:-1]) < time_limit]
+        return list(filter(lambda e: e in str(filtered_service_accounts), vault_account_tuple))
+
+    @filterService(name='vaultopenqa')
+    def parser_service_account_list(self, iam_service):
```
asmorodskyi, line 272:
```suggestion
    def get_service_accounts(self, iam_service):
```

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
...
+        while True:
+            resp = req.execute()
+            service_accounts_emails += [service_account['email']
+                                        for service_account in resp.get('accounts', [])]
+
+            req = iam_service.projects().serviceAccounts().list_next(
+                previous_request=req, previous_response=resp)
+            if req is None:
+                break
+        return tuple(set(service_accounts_emails))
```
asmorodskyi, line 301: can you elaborate why collecting everything in simple array but returning set wrapped in tuple ?

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
...
                     for w in response["warnings"]:
                         self.log_warn(w["message"])
+        self.cleanup_vaultopenqa_serviceaccounts()
+
+    def cleanup_vaultopenqa_serviceaccounts(self):
+        self.log_info(f"## GCE Vault key cleaning - {self.__project} ##")
+
+        service = googleapiclient.discovery.build('iam', 'v1', credentials=self.credentials)
+        # Get service account emails
+        service_accounts_emails = self.parser_service_account_list(service)
```
asmorodskyi, line 225:
```suggestion
        service_accounts_filtered_by_email = self.parser_service_account_list(service)
```

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
...
+        # For each email found return the ones with associated old date
+        service_accounts_emails_filtered = self.filter_email_service_accounts(service, service_accounts_emails)
+        self.delete_vaultopenqa_service_accounts(service, service_accounts_emails_filtered)
+
+    def filter_email_service_accounts(self, service, vault_account_tuple, time_limit=24):
```
asmorodskyi, line 230:
```suggestion
    def filter_by_time(self, service, service_account_list, time_limit=24):
```

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
...
+            service_accounts_emails += [service_account['email']
+                                        for service_account in resp.get('accounts', [])]
+
+            req = iam_service.projects().serviceAccounts().list_next(
+                previous_request=req, previous_response=resp)
+            if req is None:
+                break
+        return tuple(set(service_accounts_emails))
+
+    def delete_vaultopenqa_service_accounts(self, iam_service, vault_accounts):
```
asmorodskyi, line 303:
```suggestion
    def delete_service_accounts(self, iam_service, vault_accounts):
```

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
...
+
+        Raises
+        ------
+        TypeError
+            This will raised when the _vault_accounts_ will not match the expected pattern in the _name_
+        HttpError
+            Http Respond Errors
+        '''
+        if (len(vault_accounts) < 1):
+            self.log_info("Nothing to delete")
```
asmorodskyi, line 329:
```suggestion
            self.log_debug("Nothing to delete")
```

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
...
+                if self.dry_run:
+                    self.log_warn(
+                        f"Deletion of vault Service Account {account_email} skipped due to dry run mode")
+                else:
+                    try:
+                        # TODO: if deletion needs key removal first
+                        # iam_service.projects().serviceAccounts().keys().delete(
+                        #     name=vault_accounts).execute()
+                        iam_service.projects().serviceAccounts().delete(
+                            name='projects/-/serviceAccounts/%s' % account_email).execute()
```
asmorodskyi, line 341:
```suggestion
                            name='projects/-/serviceAccounts/{}'.format(account_email)).execute()
```

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
...
+        HttpError
+            Http Respond Errors
+        '''
+        if (len(vault_accounts) < 1):
+            self.log_info("Nothing to delete")
+        else:
+            for account_email in list(vault_accounts):
+                if self.dry_run:
+                    self.log_warn(
+                        f"Deletion of vault Service Account {account_email} skipped due to dry run mode")
```
asmorodskyi, line 334:
```suggestion
                        "Deletion of vault Service Account {} skipped due to dry run mode".format(account_email))
```

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
...
+                else:
+                    try:
+                        # TODO: if deletion needs key removal first
+                        # iam_service.projects().serviceAccounts().keys().delete(
+                        #     name=vault_accounts).execute()
+                        iam_service.projects().serviceAccounts().delete(
+                            name='projects/-/serviceAccounts/%s' % account_email).execute()
+                    except (TypeError, HttpError) as err:
+                        self.log_err(f"Fail to delete Service Account {account_email} \n{err}")
```
asmorodskyi, line 343:
```suggestion
                        self.log_err("Fail to delete Service Account {} \n{}".format(account_email,err))
```

ocw/lib/gce.py
```diff
@@ -207,3 +215,129 @@ def cleanup_all(self):
...
+
+        while True:
+            resp = req.execute()
+            service_accounts_emails += [service_account['email']
+                                        for service_account in resp.get('accounts', [])]
+
+            req = iam_service.projects().serviceAccounts().list_next(
+                previous_request=req, previous_response=resp)
+            if req is None:
+                break
```
asmorodskyi, line 300: since very next line after this break is return anyway we can simply do this 
```suggestion
                return service_accounts_emails
```

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: *this change is ready for review*

ocw/lib/gce.py
b10n1k: not a big thing or difference. i thought it was the appropriate data structure because:
- immutability
- iteration is faster
- less memory consumption

ocw/lib/gce.py
b10n1k: IMO the whole log_* has wrong application here. i was so tempted to changed it but it is far from the scope of this task.
_debug_ has higher level than _info_. meaning that if the logging was used properly and the logging level was _info_ the debug messages would not showed up. Something that in this particular case it would be wrong as you want to have the information in the regular logs.

ocw/lib/gce.py
b10n1k: i will take a look on the other modules. As for now i have no idea.

asmorodskyi: ok

tests/test_gce.py
```diff
@@ -58,6 +58,26 @@ def delete(self, *args, **kwargs):
...
 
 
+class FakeServiceAccounts:
+
+    def __init__(self, responses):
+        self.deleted = list()
+        self.responses = responses
+
+    def list(self, *args, **kwargs):
+        print(type(self.responses))
```
asmorodskyi, line 68: leftover ?

b10n1k: yes. removed. *ready for review*

asmorodskyi (approved):
LGTM

b10n1k:
@asmorodskyi i added a small documentation in REAME. can you take another look?

asmorodskyi (changes requested):
I don't think that we need such info in PCW README . I think this file should describe only things specific to PCW . And this is purely documentation of Google Cloud API . For Google we also deleting instances . For Azure we deleting resource groups , for AWS we deleting images , disk , snapshots , VPC's , networks . Do you really thing that PCW documentation is the right place to describe how to deal with all this types , how to authenticate with AWS and Azure ?

b10n1k:
> I don't think that we need such info in PCW README . I think this file should describe only things specific to PCW . And this is purely documentation of Google Cloud API . For Google we also deleting instances . For Azure we deleting resource groups , for AWS we deleting images , disk , snapshots , VPC's , networks . Do you really thing that PCW documentation is the right place to describe how to deal with all this types , how to authenticate with AWS and Azure ?

My opinion is that this is a good place because it has the cleanup and it can have the relative information in one place. If not in README maybe in the wiki. But i can remove this from the README now and take a look later

asmorodskyi:
Due to changes happened outside of pcw ( drop of vault ) this functionality is not needed anymore hence PR will be closed .

PR #127, opened 2021-12-15
Refactor logging for Providers using LoggerAdapter

b10n1k:
Replace hardcoding functions with the LoggerAdapter provided from logging module.

The benefits are that you leverage the logging module which can give more flexibility and there is no duplication.

Signed-off-by: ybonatakis <ybonatakis@suse.com>

### DIFF
--- a/ocw/lib/EC2.py
+++ b/ocw/lib/EC2.py
@@ -49,8 +49,11 @@ class EC2(Provider):
                 self.get_all_regions()
                 return True
             except Exception:
-                self.log_info("check_credentials (attemp:{}) with key {} expiring at {} ", i, self.__key,
-                              self.__credentials.getAuthExpire())
+                self.loggerAdapter.info(
+                    "check_credentials (attemp:{}) with key {} expiring at {} ".
+                    format(i,
+                           self.__key,
+                           self.__credentials.getAuthExpire()))
                 time.sleep(1)
         self.get_all_regions()
 
@@ -102,17 +105,18 @@ class EC2(Provider):
             response['Snapshots'].sort(key=lambda snapshot: snapshot['StartTime'].timestamp())
             for snapshot in response['Snapshots']:
                 if EC2.needs_to_delete_snapshot(snapshot, cleanup_ec2_max_snapshot_age_days):
-                    self.log_info("Deleting snapshot {} in region {} with StartTime={}", snapshot['SnapshotId'],
-                                  region, snapshot['StartTime'])
+                    self.loggerAdapter.info(
+                        "Deleting snapshot {} in region {} with StartTime={}".format(snapshot['SnapshotId'],
+                                                                                     region, snapshot['StartTime']))
                     try:
                         if self.dry_run:
-                            self.log_info("Snapshot deletion of {} skipped due to dry run mode",
-                                          snapshot['SnapshotId'])
+                            self.loggerAdapter.info(
+                                "Snapshot deletion of {} skipped due to dry run mode".format(snapshot['SnapshotId']))
                         else:
                             self.ec2_client(region).delete_snapshot(SnapshotId=snapshot['SnapshotId'])
                     except ClientError as ex:
                         if ex.response['Error']['Code'] == 'InvalidSnapshot.InUse':
-                            self.log_info(ex.response['Error']['Message'])
+                            self.loggerAdapter.info(ex.response['Error']['Message'])
                         else:
                             raise ex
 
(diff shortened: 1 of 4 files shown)


asmorodskyi (changes requested):
sorry I disagree . you basically reverting to state which we already had . I like current state more . You speaking about "flexibility" but I haven't seen that you actually using it anywhere . Also I don't see any code duplication in current version

b10n1k:
> sorry I disagree . you basically reverting to state which we already had . I like current state more . You speaking about "flexibility" but I haven't seen that you actually using it anywhere . Also I don't see any code duplication in current version


Sorry i didnt know how it was before. 
with _flexibility_ i refer to the fact that you do not need to use _custom functions_. you remove 5 functions, which it seems like `reinventing the wheel`. _flexibility_ also because you have now a dedicated class  to modify the Log (this is one place instead of 5). And lastly you could just reuse the class in any other place with just creating an instance. Removes the dependency on providers.py.

i just wanted to present my idea.  But if the code was like that before what was the problem to go to the current implementation?

asmorodskyi:
after think about this idea second time I like it . But we need to work on implementation details ... ( see my comments inline )

ocw/lib/EC2.py
```diff
@@ -49,8 +49,11 @@ def check_credentials(self):
                 self.get_all_regions()
                 return True
             except Exception:
-                self.log_info("check_credentials (attemp:{}) with key {} expiring at {} ", i, self.__key,
-                              self.__credentials.getAuthExpire())
+                self.loggerAdapter.info(
```
asmorodskyi, line 52:
```
self.loggerAdapter.info("check_credentials (attemp:{}) with key {} expiring at {} ", i, self.__key, self.__credentials.getAuthExpire()))
```

in  such a case use of format is not necessary . Logger functions may handle formatting internally . This comments applies to all other lines where you are using format within debug/info/warning/error

ocw/lib/provider.py
```diff
@@ -7,12 +7,22 @@
...
+    def process(self, msg, kwargs):
+        return '[{}] {}'.format(self.extra['ns'], msg), kwargs
+
+
 class Provider:
 
     def __init__(self, namespace: str):
         self._namespace = namespace
         self.dry_run = PCWConfig.getBoolean('default/dry_run')
         self.logger = logging.getLogger(self.__module__)
```
asmorodskyi, line 24:
```suggestion
        self.logger = ProviderAdapter(logging.getLogger(self.__module__), {'ns': namespace}) 
```

no point to have two loggers let's have one

ocw/lib/provider.py
```diff
@@ -7,12 +7,22 @@
 import logging
 
 
+class ProviderAdapter(logging.LoggerAdapter):
```
asmorodskyi, line 10: this class would make sense if you will use it not only for provider's ancestors . please replace all other loggers with it . also in such a case it would be not right to have it together with provider . please move it to separate file

asmorodskyi:
> after think about this idea second time I like it . But we need to work on implementation details ... ( see my comments inline )

ok I change it back :) Let me try to explain : if we speaking about Provider and it's ancestors only I don't see any difference between current approach and what you suggesting . it is just two different ways to do same thing .  

> with flexibility i refer to the fact that you do not need to use custom functions. 

ehm it is really hard to comment on this . we have different understanding of what "flexibility" means . In my understanding "flexibility" is ability to change something easily . How is ability to change something is blocked by use of custom functions ? and is use of custom functions is always unconditional evil ?

>  flexibility also because you have now a dedicated class to modify the Log (this is one place instead of 5)

Well I see this a bit different . I see that I have **single** place to deal with logging it is called Provider class . Having this outside Provider class would really bring a value if we will be able to reuse it anywhere . And this was actually why I wrote my previous message . I thought ok if we will reuse this class in other places it would really help . But than I goes through all places where we actually have Logger's and **none of them able to use your class** . 

> Removes the dependency on providers.py.

and what problem does it solving ? Keep in mind that it is used only in ancestors of Provider which still have dependency on providers.py anyway 

> But if the code was like that before what was the problem to go to the current implementation? 
the problem that `[namespace]` was added manually on every call of logger before . Yes your approach solving that  and saving few lines within Provider class . But in implementation I like this : 

`self.log_info("check_credentials (attemp:{}) with key {} expiring at {} ", i, self.__key, self.__credentials.getAuthExpire())`

more than this : 

`self.loggerAdapter.info( "check_credentials (attemp:{}) with key {} expiring at {} ".format(i, self.__key, self.__credentials.getAuthExpire()))`

ocw/lib/EC2.py
asmorodskyi: please ignore this

ocw/lib/provider.py
asmorodskyi: please ignore this

ocw/lib/provider.py
asmorodskyi: please ignore this

b10n1k:
rejected.

PR #131, opened 2022-01-24
Add /health.json and /instances.json

grisu48:
Add `/instances.json` path to query all currently known instances as json object and `/health.json` to query the health of the instance. Both are required for CLI tools to query the instance without using the WebUI.

This PR addresses https://github.com/SUSE/pcw/issues/124

### DIFF
--- a/ocw/urls.py
+++ b/ocw/urls.py
@@ -3,7 +3,9 @@ from . import views
 
 urlpatterns = [
     path('', views.FilteredInstanceTableView.as_view(), name='instances'),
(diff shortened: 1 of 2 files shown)


grisu48:
Codecod complains because I introduced new http methods in a class, which is also not covered by any other tests before.

grisu48:
Poking @asmorodskyi again to review this PR.

asmorodskyi:
cool idea and nice implementation :+1:

PR #133, opened 2022-03-09
Add Dockerfile and improve README

grisu48:
Adds a example `Dockerfile` and improves the run and build part of the README.
Improvements includes updates to match the current state of Publiccloud-Watcher.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM registry.opensuse.org/opensuse/leap:15.3
+
+ENV PYTHONDONTWRITEBYTECODE=1 PYTHONUNBUFFERED=1 UWSGI_WSGI_FILE=/pcw/webui/wsgi.py UWSGI_MASTER=1
+ENV UWSGI_HTTP_AUTO_CHUNKED=1 UWSGI_HTTP_KEEPALIVE=1 UWSGI_LAZY_APPS=1 UWSGI_WSGI_ENV_BEHAVIOR=holy
+
+## System preparation steps ################################################# ##
+
+COPY . /pcw/
+
+# We do the whole installation and configuration in one layer:
+# * Install system requirements
+RUN zypper -n ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_15.3/ SUSE_CA && \
+    zypper -n in ca-certificates-suse python3 python3-devel python3-pip gcc && rm -rf /var/cache
+# Install pip requirements
+RUN pip install -r /pcw/requirements.txt && rm -rf /var/cache
+
+## Finalize ################################################################# ##
+
+VOLUME /etc/pcw.ini
+VOLUME /pcw/db
+
+EXPOSE 8000/tcp
+
+WORKDIR /pcw
+# Once we are certain that this runs nicely, replace this with ENTRYPOINT.
+CMD ["/pcw/container-startup", "run"]
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+# Default container tag
+CONT_TAG=suse/qac/pcw
+
 .PHONY: all
 all: prepare test
 
@@ -15,3 +18,10 @@ test:
 .PHONY: codecov
 codecov:
 	pytest -v --cov=./ --cov-report=html && xdg-open htmlcov/index.html
+
+# Build containers
+docker-container:
+	docker build . -t ${CONT_TAG}
+podman-container:
+	podman build . -t ${CONT_TAG}
+
--- a/README.md
+++ b/README.md
@@ -3,54 +3,83 @@
 
 # OpenQA Public cloud Helper
 
-A web app which monitors, displays and deletes CSPs and linked resources. It tries to identify
-left over instances and also performs some cleanup task. For instance it deletes old custom
-uploaded images. Behavior differs per CSP.
+PublicCloud-Watcher or `pcw` is a web app which monitors, displays and deletes resources on various Cloud Service Providers (CSPs). It identifies abandoned instances by searching for certain tags and performs cleanup tasks. It also deletes old custom uploaded images. The behavior differs per CSP.
 
-## Installation
+## Install
 
-```
-virtualenv venv
-. venv/bin/activate
-pip install -r requirements.txt
-```
+See the [requirements.txt](requirements.txt). It's recommended to setup `pcw` in a virtual environment to avoid package collisions:
 
-## Requirements
+    virtualenv venv
+    . venv/bin/activate
+    pip install -r requirements.txt
 
- [Listed in requirements.txt](requirements.txt)
+## Configure and run
 
+Configuration of Publiccloud-Watcher happens via a global config file in `/etc/pcw.ini`. See [templates/pcw.ini](templates/pcw.ini) for a configuration template. To start, copy the template over:
 
-## Run django webui
+    cp templates/pwc.ini /etc/pcw.ini
 
-First copy the [pwc.ini](templates/pcw.ini) to _/etc_
+The bare minimum configuration requires the `vault[user]` and `vault[password]` settings.
 
-```
-cp templates/pwc.ini /etc/pwc.ini
-```
+PCW supports email notifications about left-over instances. See the `notify` section therein and their corresponding comments.
 
-Open and edit _vault[user]_ and _vault[password]_. Those are required.
-Add _notify[to]_ and _notify.namespace.qac[to]_ in case you want to receive notifications.
-
-```
+```bash
+# Setup virtual environment
 virtualenv env
 source env/bin/activate
 pip install -r requirements.txt
 
+
+## Configuration steps, only required once to setup the database and user
+# Setup database
 python manage.py migrate
+# Setup superuser
 python manage.py createsuperuser --email admin@example.com --username admin
 python manage.py collectstatic
+
+
+## Running the webapp server
 python manage.py runserver
 ```
-=> http://127.0.0.1:8000/
+
+By default, PCW runs on http://127.0.0.1:8000/
+
+## Building a container
+
+To build a docker/podman container with the default `suse/qac/pcw` tag, run
+
+    make docker-container
+    make podman-container
+
+This repository contains the skeleton `Dockerfile` for building a Publiccloud-Watcher docker/podman container.
+
+The resulting container requires the `/data` volume to be mounted which shall contain the `pcw.ini` file. Therein the database will also be stored in `/data/db`.
+
+The container expects the `/etc/pcw.ini` file and the `/pcw/db` directory to be mounted. To create a container using e.g. the data directory `/srv/pcw` for both volumes and expose port 8000, run the following:
+
+    podman create --hostname pcw --name pcw -v /srv/pcw/pcw.ini:/etc/pcw.ini -v /srv/pcw/db:/pcw/db -p 8000:8000/tcp suse/qac/pcw
+    podman start pcw
+
+For usage in docker simply replace `podman` by `docker` in the above command.
+
+The `pcw` container runs by default the `/pcw/container-startup` startup helper script. You can interact with it by running
+
+    podman exec pcw /pcw/container-startup help
+    
+    podman run -ti --rm --hostname pcw --name pcw -v /srv/pcw/pcw.ini:/etc/pcw.ini -v /srv/pcw/db:/pcw/db -p 8000:8000/tcp suse/qac/pcw /pcw/container-startup help
+
+To create the admin superuser within the created container named `pcw`, run
+
+    podman run -ti --rm -v /srv/pcw/pcw.ini:/etc/pcw.ini -v /srv/pcw/db:/pcw/db -p 8000:8000/tcp suse/qac/pcw /pcw/container-startup createsuperuser --email admin@example.com --username admin
 
 ## Codecov
 
 Running codecov locally require installation of `pytest pytest-cov codecov`.
 Then you can run it with
-```
-BROWSER=$(xdg-settings get default-web-browser)
-pytest -v --cov=./ --cov-report=html && $BROWSER htmlcov/index.html
-```
+
+    BROWSER=$(xdg-settings get default-web-browser)
+    pytest -v --cov=./ --cov-report=html && $BROWSER htmlcov/index.html
+
 and explore the results in your browser
 
 ## Debug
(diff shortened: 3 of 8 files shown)


Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM registry.opensuse.org/opensuse/leap:latest
+
+ENV PYTHONDONTWRITEBYTECODE=1 PYTHONUNBUFFERED=1 UWSGI_WSGI_FILE=/pcw/webui/wsgi.py UWSGI_MASTER=1
+ENV UWSGI_HTTP_AUTO_CHUNKED=1 UWSGI_HTTP_KEEPALIVE=1 UWSGI_LAZY_APPS=1 UWSGI_WSGI_ENV_BEHAVIOR=holy
+
+COPY . /pcw/
+WORKDIR /pcw
```
b10n1k, line 7: Those lines better to go later as they more likely to change than the zypper cmds. wdyt?

pdostal: I agree here with Yianni. I suggest:
 1. Run `zypper`
 2. Copy `requirements.txt`
 3. Run `pip`
 4. Copy the rest.

README.md
```diff
@@ -3,54 +3,72 @@
 
 # OpenQA Public cloud Helper
 
-A web app which monitors, displays and deletes CSPs and linked resources. It tries to identify
-left over instances and also performs some cleanup task. For instance it deletes old custom
-uploaded images. Behavior differs per CSP.
+Publiccloud-Watcher or pcw is a a web app which monitors, displays and deletes resources on various cloud service providers (CSPs) and linked resources. It identifies abandoned instances by searching for certain tags and performs cleanup tasks. It also deletes old custom uploaded images. Behavior differs per CSP.
```
jlausuch, line 6:
```suggestion
PublicCloud-Watcher or pcw is a web app which monitors, displays and deletes resources on various Cloud Service Providers (CSPs). It identifies abandoned instances by searching for certain tags and performs cleanup tasks. It also deletes old custom uploaded images. The behavior differs per CSP.
```
not sure what you mean about `... and linked resources`

README.md
```diff
@@ -3,54 +3,72 @@
...
+## Building a container
+
+To build a docker/podman container with the default `suse/qac/pcw` tag, run
+
+    make docker-container
+    make podman-container
+
+This repository contains the skeleton `Dockerfile` for building a Publiccloud-Watcher docker/podman container.
+
+The resulting container requires the `/conf` volume to be mounted and contain the `pcw.ini` file.
```
jlausuch, line 56:
```suggestion
The resulting container requires the `/conf` volume to be mounted which shall contain the `pcw.ini` file.
```

README.md
```diff
@@ -3,54 +3,72 @@
...
+
+To run the resulting container using e.g. the config directory `/srv/pcw` (must contain `pcw.ini`!), the corresponding environment variables and publish port 8000, do
+
+    podman run -ti --rm --hostname pcw --name pcw -v /srv/pcw:/conf -p 8000:8000/tcp suse/qac/pcw
+
+For usage in docker simply replace `podman` with `docker` in the above command.
```
jlausuch, line 62:
```suggestion
For usage in docker simply replace `podman` by `docker` in the above command.
```

README.md
```diff
@@ -61,5 +79,5 @@ To simplify problem investigation pcw has two [django commands](https://docs.dja
 
 [updaterun](ocw/management/commands/updaterun.py)
 
-this allows triggering core functionality without web UI. It is highly recommended to use `dry_run = True` in `pcw.ini` in
+those allows triggering core functionality without web UI. It is highly recommended to use `dry_run = True` in `pcw.ini` in
```
jlausuch, line 82:
```suggestion
those allow triggering core functionality without web UI. It is highly recommended to use `dry_run = True` in `pcw.ini` in
```

jlausuch (commented):
Always happy to see better documentation :)
Just some nitpicking comments and suggestions.

README.md
grisu48: Yeah, that's too vague, especially for the first paragraph. I removed it as you suggested, that reads nice.

Dockerfile
grisu48: I updated the Dockerfile accordingly, however it is highly unlikely that any package interferes with the custom `/pcw` directory.

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+# Install system requirements
+RUN zypper -n in python3 python3-devel python3-pip gcc
+## Install SUSE CA. Comment the following two lines, if this is not desired
+RUN zypper -n ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_15.3/ SUSE_CA
+RUN zypper -n in ca-certificates-suse
+COPY requirements.txt /
+# Install pip requirements
+RUN pip install -r requirements.txt
+# The cache (especially zypper) is not required for the resulting container
+RUN rm -rf /var/cache /requirements.txt
```
pdostal, line 17: This is useless and should be done after each `zypper in` call directly (as the same `RUN`

Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM registry.opensuse.org/opensuse/leap:latest
```
asmorodskyi, line 1: can you elaborate reasons for choosing such base ? there is well maintained python3 docker images . don't see good reason to go one level deeper and take this burden on our shelters ...

grisu48: This happened for two reasons: First, the `python3` container had some linker issues, which could be solved by using the python38 or python36 container (I don't remember which one worked). Second: I believe `pcw` would be a prime candidate to use our new and shiny BCI containers, once they are available. To make this transition happen as smooth as possible, I believe switching to leap could be a good step towards this direction.

asmorodskyi: ok make sense

asmorodskyi, line 1: but let's stick to some certain version here ( 15.3 )

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN ln -s /pcw/ocw/static /pcw/static
+
+## Finalize ################################################################# ##
+
+VOLUME /data
+EXPOSE 8000/tcp
+
+WORKDIR /pcw
+# Once we are certain that this runs nicely, replace this with ENTRYPOINT.
+CMD ["/pcw/pcw", "run"]
```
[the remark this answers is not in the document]
grisu48: See lines 28-29 in the Dockerfile

    # Create /pcw/pcw for convenience
    RUN ln /pcw/container-startup /pcw/pcw

The reason to link this is that if one needs to run custom `pcw` commands (e.g. add admin emails, migrate ...) this can be done with a nice-to-remember path, instead of performing the whole `python migrate.py runserver` magic.

Dockerfile
grisu48: Ok

container-startup
```diff
@@ -0,0 +1,51 @@
...
+  	;;
+  "--help")
+  	usage
+  	exit 0
+  	;;
+  "run")
+    # Ensure the database directory exists
+    mkdir -p /data/db
+  	python3 manage.py migrate
+  	python3 manage.py runserver 0.0.0.0:8000
```
asmorodskyi, line 41: have you read `DO NOT USE THIS SERVER IN A PRODUCTION SETTING. ` in description of runserver command ?

container-startup
```diff
@@ -0,0 +1,51 @@
...
+  	usage
+  	exit 0
+  	;;
+  "--help")
+  	usage
+  	exit 0
+  	;;
+  "run")
+    # Ensure the database directory exists
+    mkdir -p /data/db
```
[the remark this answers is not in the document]
grisu48: `createsuperuser` can be executed via `podman exec`. If this is necessary, it will be documented in the publiccloud-qa-suse-de repository. This command is the reason, why I created `/pcw/pcw` in the container to make it easier to execute such commands (previous command).

I have no idea what `collectstatic` does and if we need it.

container-startup
grisu48: Yes. This server runs within a container, so I don't see any other generic way of permitting access to the django webserver other than allowing any host. Since the container provides some isolation itself, this should also be ok to have as default.

If you have however an idea how to make this better, please go ahead :slightly_smiling_face:

Dockerfile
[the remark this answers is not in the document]
grisu48: This also allows to easily call additional commands, e.g. `createsuperuser` (see the next comment).

container-startup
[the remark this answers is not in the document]
asmorodskyi: > createsuperuser can be executed via podman exec

why not to keep everything in one place as it was before ? 

> I have no idea what collectstatic does and if we need it.

we need to clarify this

container-startup
asmorodskyi: I am speaking about something different here . In vault-docker we not using `manage.py runserver` at all to start pcw container . instead we doing it with `uwsgi --enable-threads --socket :3000 --module webui.wsgi`

grisu48:
I updated the `Dockerfile` and the `container-startup` with the suggestions from @pdostal and @asmorodskyi regarding two things:

1. Remove unnecessary layers and crop the Dockerfile down to one `COPY` and one `RUN`, so the resulting container does not consist of lots of layers
2. Usage of `uwsgi` in order to run the django webapp instead of running `runserver` via python.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+COPY requirements.txt /requirements.txt
+RUN pip install -r /pcw/requirements.txt && rm -rf /var/cache
+# * Create links to config files and database directory on the /data volume
+# * Create /pcw/pcw for convenience
+# * pcw assumes the static content to be located in /pcw/static
+RUN ln -s /data/pcw.ini /etc/pcw.ini && \
+    ln -s /data/db /pcw/db && \
+    ln /pcw/container-startup /pcw/pcw && \
+    ln -s /pcw/ocw/static /pcw/static && \
+    rm -rf /var/cache /pcw/requirements.txt
```
pdostal, line 25: No need for this.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    zypper -n ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_15.3/ SUSE_CA && \
+    zypper -n in ca-certificates-suse && rm -rf /var/cache
+# Install pip requirements
+COPY requirements.txt /requirements.txt
+RUN pip install -r /pcw/requirements.txt && rm -rf /var/cache
+# * Create links to config files and database directory on the /data volume
+# * Create /pcw/pcw for convenience
+# * pcw assumes the static content to be located in /pcw/static
+RUN ln -s /data/pcw.ini /etc/pcw.ini && \
+    ln -s /data/db /pcw/db && \
```
pdostal, line 22: Just use `/data` and you don't need to link stuff. I just find this useless.

container-startup
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash -e
```
pdostal, line 1: There is in my opinion no need for this script. As you said, we can simply use `podman run python3 manage.py migrate` and so on.

container-startup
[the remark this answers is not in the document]
pdostal: We in my opinion simply don't have to do the migrations every time.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+COPY . /pcw/
+
+# We do the whole installation and configuration in one layer:
+# * Install system requirements
+RUN zypper -n in python3 python3-devel python3-pip gcc && \
+    zypper -n ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_15.3/ SUSE_CA && \
+    zypper -n in ca-certificates-suse && rm -rf /var/cache
+# Install pip requirements
+COPY requirements.txt /requirements.txt
```
asmorodskyi, line 16: you already copying everything in line 8 so maybe this can be omitted ?

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+## System preparation steps ################################################# ##
+
+COPY . /pcw/
+
+# We do the whole installation and configuration in one layer:
+# * Install system requirements
+RUN zypper -n in python3 python3-devel python3-pip gcc && \
+    zypper -n ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_15.3/ SUSE_CA && \
+    zypper -n in ca-certificates-suse && rm -rf /var/cache
```
asmorodskyi, line 14: as you already have `.dockerignore` no need to do any manual cleanups ?

asmorodskyi: also if you will first add repo and than do install call you may put all packages in one zypper call

Makefile
```diff
@@ -15,3 +18,10 @@ test:
 .PHONY: codecov
 codecov:
 	pytest -v --cov=./ --cov-report=html && xdg-open htmlcov/index.html
+
+# Build containers
+docker-container:
+	docker build . -t ${CONT_TAG}
```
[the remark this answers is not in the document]
asmorodskyi: my personal opinion no need to add things for "just in case " . but it is just IMHO here . feel free to keep it . resolving the thread

container-startup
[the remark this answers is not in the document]
asmorodskyi: not sure what do you mean when you saying "every time" . main idea which we need to keep in mind - this approach suppose to work out-of-the-box on clean machine not having anything . Our current state in `vault-docker` project there is no need to run any commands outside of bash scripts . I don't care if we will migrate this into ansible/salt/you-name-it but main thing which I would like to keep - I fire one line in terminal and than some magic install everything what I need and starting all needed containers . And this command suppose to work on fresh installed system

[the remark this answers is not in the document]
asmorodskyi: if instead we will have approach where there will be well documented flow which force me to execute set of commands one by one I would treat that as step back

Dockerfile
grisu48: Removed.

container-startup
grisu48: The main reason for this script is that we need to run a sequence of commands at startup. Please check `pcw/docker-entrypoint.sh` in the old `vault-docker` repository if you don't believe me.

grisu48: I also cannot comment if this is needed on every restart. It was done so before and I simply adapted the current workflow without major modifications, that's why it's there.

Dockerfile
grisu48: I merged the `zypper` calls together, however the `rm -rf /var/cache` is still required, because that's outside the `/pcw` directory and those contains the system (and zypper) cache and thus nothing we can control via `.dockerignore`.

pdostal (approved):
OK for me, thank you Felix! :rocket:

PR #141, opened 2022-05-17
switch from Vault to local cached creds

asmorodskyi:


asmorodskyi:
closing , will reopen #139

PR #143, opened 2022-05-24
Switch container workflow to GitHub Packages

grisu48:
This commit switches the container CD to use the internal GitHub
registry by changing two things:

* Change the Dockerfile back to use Leap and pip for dependency
  installing
* Add Github workflow for building and releasing new containers whenever
  we make a new pcw release

This means that every new release builds a new container directly in
this repository, so we do not rely on any other service (OBS or
Dockerhub) for the pcw container.


grisu48:
Run failure unrelated to this PR.

pdostal (approved):
Please rebase so the is #145 gets affected.

PR #148, opened 2022-06-14
Expose regions in EC2 into config

asmorodskyi:


cfconrad (approved):
LGTM
I think there is no easy better way to do it.
As we do not have `list` in ome of the `get_feature_property()` return values. And even the default should be something computer generated, as we ask `aws.ec2` for the full list...

But I would consider to add it in the example configuration in https://github.com/SUSE/pcw/blob/master/templates/pcw.ini#L1

PR #152, opened 2022-07-12
Extend README with what actually PCW is doing

asmorodskyi:


README.md
```diff
@@ -3,9 +3,35 @@
 
 # OpenQA Public cloud Helper
 
-PublicCloud-Watcher or `pcw` is a web app which monitors, displays and deletes resources on various Cloud Service Providers (CSPs). It identifies abandoned instances by searching for certain tags and performs cleanup tasks. It also deletes old custom uploaded images. The behavior differs per CSP.
-
-The fastest way to run PublicCloud-Watcher is via the provided containers, as described in the [Running a container](#running-a-container) section.
+PublicCloud-Watcher (PCW) is a web app which monitors, displays and deletes resources on various Cloud Service Providers (CSPs).
+PCW has two main flows :
+
+1. Update run ( implemented in [ocw/lib/db.py](ocw/lib/db.py) )
```
grisu48, line 9:
```suggestion
1. Periodic update runs (implemented in [ocw/lib/db.py](ocw/lib/db.py))
```

README.md
```diff
@@ -3,9 +3,35 @@
...
+ Concentrates on deleting VMs (in case of Azure Resource Groups) . Each update run scans accounts defined in configuration file and
```
grisu48, line 10: Why does an Update run concentrates on deleting VMs?

README.md
```diff
@@ -3,9 +3,35 @@
...
+and persists all results of each scan into sqlite database. If it is new entity than it obligatory gets TTL. TTL may be taken from tag `openqa_ttl` if
```
grisu48, line 11: duplicate and (see line before).

```suggestion
writes the obtained results into a local sqlite database. Newly discovered entities get assigned an obligatory time-to-life value (TTL). TTL may be taken from tag `openqa_ttl` if
```

Also is "obligatory" the right word or do you mean "default"? Obligatory means every entity gets a TTL value, but as we will see later, one can bypass with by setting `pcw_ignore`.

README.md
```diff
@@ -3,9 +3,35 @@
...
+entity is tagged with such tag if not PCW will check `pcw.ini` for `updaterun/default_ttl` setting if setting is not defined than PCW will use hard-coded value from [webui/settings.py](webui/settings.py). Database has a web UI where you can manually trigger certain entity deletion.
+
+ After persisting results into db PCW deciding which entities needs to be deleted. There are two ways to survive for entity :
```
grisu48, line 14: If the "Update" run also deletes CSP resources, I would actually rename it so "Seek for leftovers" or something similar. Update suggests no destructive means are taken place.

README.md
```diff
@@ -3,9 +3,35 @@
...
+   a. Having tag `pcw_ignore` ( with any value)
+   b. Age of entity is lower than TTL defined. Age is calculated as delta of last_seen and first_seen
+
+  For entities that survive cleanup PCW will sent notification email to the list defined in config.
```
grisu48, line 18: Is this true? If I check the email sent by PCW then I get "leftover" emails, i.e. instances that would be deleted. So it is the opposite, no?

README.md
```diff
@@ -3,9 +3,35 @@
...
+
+2. Cleanup ( implemented in [ocw/lib/cleanup.py](ocw/lib/cleanup.py) )
+ Concentrates on everything except VM deletion . This vary a lot per CSP so let's clarify that on per provider level.
```
grisu48, line 21: "Concentrates on everything except VM deletion" is only useful for people who know already the internals and know what "everything except VM deletion" means. I have no idea what that means, but based on the text below I assume it means the deletion of dangling resources.

```suggestion
 Concentrates on deleting dangling resources from the previously scannerd virtual machines, e.g. disk images. The actual behaviour varies per CSP:
```

grisu48 (commented):
Thank you Anton, this is very nice! I left a couple of comments and tried to give you concrete suggestions whenever possible. Some places I didn't quiet understood and would request a clarification there, but in overall I really like the improved documentation and can't wait to see this merged! :+1:

README.md
asmorodskyi: I name this section like this because it coming from main function which handle it https://github.com/SUSE/pcw/blob/master/ocw/lib/db.py#L190 . Why this function named `update_run` no idea :) . I am fine to rename a function to any other name which you think suits better but I see this idea not directly related to current PR . In separate PR we may also update section name accordingly

README.md
asmorodskyi: I changed my mind :) see details https://github.com/SUSE/pcw/pull/152#discussion_r918938877

README.md
asmorodskyi: suggestion accepted but I will section open due to your concerns with "obligatory"

README.md
asmorodskyi: sure make sense but I would go with function renaming too and move this into dedicated PR

README.md
asmorodskyi: you not getting email with entities to be deleted . Idea of this email to highlight what was NOT deleted so you as a human can jump in if it is still make sense to delete it

README.md
asmorodskyi: ehm but to know this you need to read few lines above . I assume I may rely on a fact that person read whole README and should not phrase text expecting every sentence to be self-containing

README.md
grisu48: Ah ok, from that perspective it makes sense.

README.md
grisu48: Not in general, but the sentence at the beginning of a paragraph should contain the most important message of that paragraph. Those sentences should describe what is going on and why the paragraph is important. The first sentence should be indeed self-containing.

This is arguably the most important/condensed piece of information and if the reader is already starved for contextual information here, then the read flow is disturbed and most readers annoyed.

README.md
grisu48: What do you mean with `I will section open due to your concerns with "obligatory"`?

README.md
asmorodskyi: it is not paragraph but point of the list which is very important difference in context of our conversation . So basically you have **point of the list** number 1 saying "**Update run ( implemented in [ocw/lib/db.py](ocw/lib/db.py) )** Concentrates on deleting VMs (in case of Azure Resource Groups)."  and point of the list number 2 saying "**Cleanup ( implemented in [ocw/lib/cleanup.py](ocw/lib/cleanup.py) )** Concentrates on everything except VM deletion.". To me the list is self-containing . I am insisting on such wording because for me it is part of whole architectural desing of PCW which has two main threads and one thread concentrates on deletion of VMs and another on **everything else** so whenever we would decide to add another entity to monitor it will gets into cleanup htread and not into update run ( or whatever new name this thread will have )

README.md
asmorodskyi: > I would still add the word `periodic` to highlight that this is a recurring task.

I addressed that but not exactly like you asked. Check latest version and tell me if it is fine

README.md
grisu48: Correct me if I'm wrong but I don't see "obligatory" adding any value in this section. Obligatory is a strong word and here I don't see it highlighting anything meaningful, or did I miss something?

I would like to suggest to remove obligatory here, but leave the last word to you. IMHO it highlights something that is not there and I find it a bit off.

README.md
grisu48: To come back to my original comment here: IMHO "Concentrates on everything except VM deletion" is too unspecific and doesn't give the reader any new information. It should be more specific on what the "everything else is", as suggested in my original comment.

I leave it however up to you to accept it or leave it as it is.

README.md
asmorodskyi: I have no special favor exactly to word "obligatory" . My point here was to underline that every VM will get TTL so basically unless you put `pcw_ingore` it will be deleted after TTL defined time . but if you think it is clear without this word too I am fine to drop it.

grisu48 (approved):

PR #153, opened 2022-07-20
Extend logging and fix missing namespace in where clause

asmorodskyi:


cfconrad (approved):
LGTM beside the discussed thing...

ocw/lib/db.py
```diff
@@ -34,18 +34,20 @@ def sync_csp_to_local_db(pc_instances, provider, namespace):
...
+            o = Instance.objects.get(provider=i.provider, instance_id=i.instance_id, vault_namespace=namespace)
             if o.region != i.region:
-                logger.info("Instance %s:%s changed region from %s to %s",
-                            provider, i.instance_id, o.region, i.region)
+                logger.info("[%s] Instance %s:%s changed region from %s to %s",
+                            namespace, provider, i.instance_id, o.region, i.region)
                 o.region = i.region
             if o.state == StateChoice.DELETED:
-                logger.error("Update already DELETED instance %s:%s\n\t%s", provider, i.instance_id, i.csp_info)
+                logger.error("[{}] Update already DELETED instance {}:{} {}".format(
```
mpagot, line 45: Why do you need to move from the lazy formatting to ''.format(...)?

ocw/lib/db.py
```diff
@@ -249,7 +251,8 @@ def auto_delete_instances():
                      age__gte=F('ttl')).exclude(csp_info__icontains='pcw_ignore')
         email_text = set()
         for i in o:
-            logger.info("[{}][{}] TTL expire for instance {}".format(i.provider, i.vault_namespace, i.instance_id))
+            logger.info("[{}][{}] TTL expire for instance {} {}".format(
```
mpagot, line 254: mind to change in a lazy formatter

ocw/lib/db.py
```diff
@@ -58,7 +60,7 @@ def sync_csp_to_local_db(pc_instances, provider, namespace):
         o.csp_info = i.csp_info
         o.last_seen = t_now
         o.active = True
-        o.age = o.last_seen - o.first_seen
+        o.age = t_now - i.first_seen
```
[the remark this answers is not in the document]
asmorodskyi: ok

ocw/lib/db.py
asmorodskyi: reverted back

ocw/lib/db.py
asmorodskyi: reverted back

ocw/lib/db.py
```diff
@@ -34,18 +34,20 @@ def sync_csp_to_local_db(pc_instances, provider, namespace):
...
-                            provider, i.instance_id, o.region, i.region)
+                logger.info("[%s] Instance %s:%s changed region from %s to %s",
+                            namespace, provider, i.instance_id, o.region, i.region)
                 o.region = i.region
             if o.state == StateChoice.DELETED:
-                logger.error("Update already DELETED instance %s:%s\n\t%s", provider, i.instance_id, i.csp_info)
+                logger.error("[%s] Update already DELETED instance %s:%s %s",
+                             namespace, provider, i.instance_id, i.all_time_fields())
             if o.state != StateChoice.DELETING:
                 o.state = StateChoice.ACTIVE
```
grisu48, line 48:
```suggestion
                logger.error("[%s] Update and reactivating DELETING instance %s:%s %s",
                             namespace, provider, i.instance_id, i.all_time_fields())
                o.state = StateChoice.ACTIVE
```

grisu48 (commented):
One logging suggestion otherwise LGTM!

If you implement my suggestion from my point of view you can proceed and merge

ocw/lib/db.py
asmorodskyi: Keep in mind that message coming just after this for same instance `logger.debug("[%s] Update instance %s:%s", namespace, provider, i.instance_id)` . While I agree with idea to mention instance activation I would suggest different message : 
`logger.error("[%s] %s:%s instance which still exists has DELETED state in DB. Reactivating %s",
                             namespace, provider, i.instance_id, i.all_time_fields())`

PR #155, opened 2022-07-22
Set correct age for instance which was already exists before

asmorodskyi:


ocw/lib/db.py
```diff
@@ -42,8 +42,9 @@ def sync_csp_to_local_db(pc_instances, provider, namespace):
                             namespace, provider, i.instance_id, o.region, i.region)
                 o.region = i.region
             if o.state == StateChoice.DELETED:
-                logger.error("[%s] %s:%s instance which still exists has DELETED state in DB. Reactivating %s",
-                             namespace, provider, i.instance_id, i.all_time_fields())
+                logger.info("[%s] %s:%s instance which still exists has DELETED state in DB. Reactivating %s",
+                            namespace, provider, i.instance_id, i.all_time_fields())
+                o.first_seen = i.first_seen
```
ilausuch, line 47: Oh, yes, this solution is better. We are maintaining the DB consistent. Good idea.

ilausuch (approved):

ocw/lib/db.py
cfconrad: Nice catch!

Maybe we should do something like:
`Delete and recreate` or init all fields, so we do not get this situation just with a other field in the future!

pdostal (commented):
Congratulations Anton

PR #161, opened 2022-09-08
Fix python3.10 tests

mimi1vx:
https://progress.opensuse.org/issues/113513

### DIFF
--- a/.github/workflows/pcw.yml
+++ b/.github/workflows/pcw.yml
@@ -11,11 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: ['3.10']
     steps:
       - uses: actions/checkout@v2
       - name: Setup Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install requirements
--- a/ocw/lib/provider.py
+++ b/ocw/lib/provider.py
@@ -3,7 +3,7 @@ import re
 from datetime import datetime
 from datetime import timedelta
 from datetime import timezone
-from distutils.version import LooseVersion
+from looseversion import LooseVersion
 import logging
 import json
 from pathlib import Path
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,10 @@
+[build-system]
+requires = ["setuptools >= 40.9.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "pcw"
+version = "1.13"
+
+[tool.setuptools]
+py-modules = []
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,3 +17,4 @@ google-api-python-client==2.55.0
 google-cloud-storage==2.4.0
 python-dateutil
 apscheduler
+looseversion
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,0 @@
-import setuptools
-
-setuptools.setup(name='pcw', version='1.0')


grisu48 (approved):
LGTM! I like it that we can get rid of some old stuff that is not needed anymore.

Would way for @asmorodskyi still for approval though before merging.

pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[build-system]
+requires = ["setuptools >= 40.9.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "pcw"
+version = "1.0"
```
pdostal, line 7: It's not `1.0` but `1.13`.
Also JFYI we're using the version number for tagging containers in [ghcr.io/SUSE/pcw](https://ghcr.io/SUSE/pcw) registry.

pdostal (approved):
I think this is cool 🚀

.github/workflows/pcw.yml
```diff
@@ -11,11 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: ['3.8', '3.10']
     steps:
       - uses: actions/checkout@v2
       - name: Setup Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
```
asmorodskyi, line 18: does it mean that we were using python2 here or what ?

.github/workflows/pcw.yml
```diff
@@ -11,11 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: ['3.8', '3.10']
     steps:
       - uses: actions/checkout@v2
```
asmorodskyi, line 16: why you not changed v2 here ?

ocw/lib/provider.py
```diff
@@ -3,7 +3,7 @@
 from datetime import datetime
 from datetime import timedelta
 from datetime import timezone
-from distutils.version import LooseVersion
+from looseversion import LooseVersion
```
asmorodskyi, line 6: this is super awesome ! I was expecting a lot of work with this and it appears just one-liner , cool !

.github/workflows/pcw.yml
```diff
@@ -11,11 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: ['3.8', '3.10']
```
asmorodskyi, line 14: I don't see any reason to keep 3.8 here , whole project already switched 3.10 . Even during testing of the code locally it is expected that people will use Docker_dev which also runs everything within python BCI container with python 3.10

.github/workflows/pcw.yml
asmorodskyi: ah it is just **package** version , right ?

pyproject.toml
mimi1vx: ahh, I got this from setup.py where was 1.0 :D

.github/workflows/pcw.yml
mimi1vx: right , its action version.

ocw/lib/provider.py
mimi1vx: unfortuqanelty new dep :(

ocw/lib/azure.py
```diff
@@ -206,7 +206,7 @@ def cleanup_images_from_rg(self, keep_images):
                     if self.dry_run:
                         self.log_info("Deletion of image {} skipped due to dry run mode", item.name)
                     else:
-                        self.compute_mgmt_client().images.delete(self.__resource_group, item.name)
+                        self.compute_mgmt_client().images.begin_delete(self.__resource_group, item.name)
```
asmorodskyi, line 209: how this appears here ?  Looks like some mess in your commits tree , this change already in master  https://github.com/SUSE/pcw/commit/e6143d526c033ef19e33516e026981a41ecba5e7

tests/test_azure.py
```diff
@@ -110,7 +110,7 @@ def mock_compute_mgmt_client(self):
         def compute_mgmt_client():
             pass
         compute_mgmt_client.images = lambda: None
-        compute_mgmt_client.images.delete = lambda rg, name: deleted_images.append(name)
+        compute_mgmt_client.images.begin_delete = lambda rg, name: deleted_images.append(name)
```
asmorodskyi, line 113: check  https://github.com/SUSE/pcw/commit/e6143d526c033ef19e33516e026981a41ecba5e7

mimi1vx: looks like something wrong with rebase ... I'll fix

PR #165, opened 2022-10-11
Expose pcw_ignore in web UI

asmorodskyi:
PR includes 3 not directly related commits : 
1. Expose pcw_ignore in web UI
2. Tweak Dockerfile_dev  allowing to run pytest inside container 
3. Decrease amount of retries on failed credentials


cfconrad (approved):
didn't run it my self.
From simply reading code LGTM!

mimi1vx (approved):
LGTM

Dockerfile_dev
```diff
@@ -6,11 +6,11 @@ ENV UWSGI_HTTP_AUTO_CHUNKED=1 UWSGI_HTTP_KEEPALIVE=1 UWSGI_LAZY_APPS=1 UWSGI_WSG
 ## System preparation steps ################################################# ##
 
 # !!! Runtime changes won't affect requirements.txt
-COPY requirements.txt /tmp/
+COPY requirements.txt requirements_test.txt /tmp/
 # * Install system requirements
 # * Install pip requirements
 # * Empty system cache to conserve some space
-RUN zypper -n in python310-devel gcc libffi-devel && pip3.10 install -r /tmp/requirements.txt && rm -rf /var/cache
+RUN zypper -n in python310-devel gcc libffi-devel && pip3.10 install -r /tmp/requirements_test.txt && rm -rf /var/cache
```
ricardobranco777, line 13: I usually add `--no-cache-dir` to `pip`.

ocw/lib/db.py
```diff
@@ -67,6 +67,15 @@ def sync_csp_to_local_db(pc_instances, provider, namespace):
         update(state=StateChoice.DELETED)
 
 
+def tag_to_boolean(tag_name, csp_info):
+    if 'tags' not in csp_info:
+        return False
+    if tag_name in csp_info['tags']:
+        return bool(csp_info['tags'][tag_name])
+    else:
+        return False
```
ricardobranco777, line 76:
```suggestion
    try:
        return bool(csp_info['tags'][tag_name])
    except KeyError:
        return False
```

ricardobranco777 (approved):
Some minor suggestions.  Otherwise LGTM.

Also didn't run it.

ocw/lib/db.py
mimi1vx: using exceptions as flow :( its expensive

mimi1vx: but for line 73-76 can be used `get`:
```
return bool(cap_info['tags'].get(tag_name, False)
```

Dockerfile_dev
asmorodskyi: ok , good to know but I assume it does not make much difference when there is `rm -rf /var/cache` , right ?

PR #166, opened 2022-10-21
Use env for SECRET_KEY

mimi1vx:


webui/settings.py
```diff
@@ -25,7 +25,7 @@
 # See https://docs.djangoproject.com/en/2.1/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
```
pdostal, line 27: What does this comment mean? cc: @asmorodskyi

mimi1vx: I bet is default comment generated by Django

PR #171, opened 2022-11-14
Redo volumes,disks,images cleanup

asmorodskyi:
Initially when data storage cleanup logic was created
we wanted to make it very flexable from one side
and very coutious from another. After more than 2 years
of practical usage of the tool we can clearly state that
half of implemented features not needed

### DIFF
--- a/ocw/lib/EC2.py
+++ b/ocw/lib/EC2.py
@@ -1,9 +1,8 @@
-from .provider import Provider, Image
+from .provider import Provider
 from webui.settings import PCWConfig, ConfigFile
 from dateutil.parser import parse
 import boto3
 from botocore.exceptions import ClientError
-import re
 from datetime import date, datetime, timedelta, timezone
 from ocw.lib.emailnotify import send_mail
 import traceback
@@ -89,32 +88,21 @@ class EC2(Provider):
         return clusters
 
     @staticmethod
-    def needs_to_delete_snapshot(snapshot, cleanup_ec2_max_snapshot_age_days) -> bool:
-        delete_older_than = date.today() - timedelta(days=cleanup_ec2_max_snapshot_age_days)
-        if datetime.date(snapshot['StartTime']) < delete_older_than:
-            regexes = [
-                re.compile(r'''^OpenQA upload image$'''),
-                re.compile(r'''^Created by CreateImage\([\w-]+\) for ami-\w+ from vol-\w+$''')
-            ]
-            for regex in regexes:
-                m = re.match(regex, snapshot['Description'].strip())
-                if m:
-                    return True
-        return False
+    def is_outdated(creation_time, valid_period_days) -> bool:
+        return datetime.date(creation_time) < (date.today() - timedelta(days=valid_period_days))
 
-    def cleanup_snapshots(self, cleanup_ec2_max_snapshot_age_days):
+    def cleanup_snapshots(self, valid_period_days):
         for region in self.all_regions:
             response = self.ec2_client(region).describe_snapshots(OwnerIds=['self'])
-            response['Snapshots'].sort(key=lambda snapshot: snapshot['StartTime'].timestamp())
             for snapshot in response['Snapshots']:
-                if EC2.needs_to_delete_snapshot(snapshot, cleanup_ec2_max_snapshot_age_days):
-                    self.log_info("Deleting snapshot {} in region {} with StartTime={}", snapshot['SnapshotId'],
-                                  region, snapshot['StartTime'])
+                if EC2.is_outdated(snapshot['StartTime'], valid_period_days):
                     try:
                         if self.dry_run:
                             self.log_info("Snapshot deletion of {} skipped due to dry run mode",
                                           snapshot['SnapshotId'])
                         else:
+                            self.log_info("Deleting snapshot {} in region {} with StartTime={}",
+                                          snapshot['SnapshotId'], region, snapshot['StartTime'])
                             self.ec2_client(region).delete_snapshot(SnapshotId=snapshot['SnapshotId'])
                     except ClientError as ex:
                         if ex.response['Error']['Code'] == 'InvalidSnapshot.InUse':
@@ -122,14 +110,13 @@ class EC2(Provider):
                         else:
                             raise ex
 
-    def cleanup_volumes(self, cleanup_ec2_max_volumes_age_days):
-        delete_older_than = date.today() - timedelta(days=cleanup_ec2_max_volumes_age_days)
+    def cleanup_volumes(self, valid_period_days):
         for region in self.all_regions:
             response = self.ec2_client(region).describe_volumes()
             for volume in response['Volumes']:
-                if datetime.date(volume['CreateTime']) < delete_older_than:
+                if EC2.is_outdated(volume['CreateTime'], valid_period_days):
                     if self.volume_protected(volume):
-                        self.log_info('Volume {} has tag DO_NOT_DELETE so protected from deletion',
+                        self.log_info('Volume {} has tag pcw_ignore so protected from deletion',
                                       volume['VolumeId'])
                     elif self.dry_run:
                         self.log_info("Volume deletion of {} skipped due to dry run mode", volume['VolumeId'])
@@ -147,7 +134,7 @@ class EC2(Provider):
     def volume_protected(self, volume):
         if 'Tags' in volume:
             for tag in volume['Tags']:
-                if tag['Key'] == 'DO_NOT_DELETE':
+                if tag['Key'] == 'pcw_ignore':
                     return True
         return False
 
@@ -209,66 +196,13 @@ class EC2(Provider):
                         self.log_info("Finally deleting {} cluster", cluster)
                         self.eks_client(region).delete_cluster(name=cluster)
 
-    def parse_image_name(self, img_name):
-        regexes = [
-            # openqa-SLES12-SP5-EC2.x86_64-0.9.1-BYOS-Build1.55.raw.xz
-            re.compile(r'''^openqa-SLES
-                              (?P<version>\d+(-SP\d+)?)
-                              -(?P<flavor>EC2)
-                              \.
-                              (?P<arch>[^-]+)
-                              -
-                              (?P<kiwi>\d+\.\d+\.\d+)
-                              -
-                              (?P<type>(BYOS|On-Demand))
-                              -Build
-                              (?P<build>\d+\.\d+)
-                              \.raw\.xz
-                              ''', re.RegexFlag.X),
-            # openqa-SLES15-SP2.x86_64-0.9.3-EC2-HVM-Build1.10.raw.xz'
-            # openqa-SLES15-SP2-BYOS.x86_64-0.9.3-EC2-HVM-Build1.10.raw.xz'
-            # openqa-SLES15-SP2.aarch64-0.9.3-EC2-HVM-Build1.49.raw.xz'
-            # openqa-SLES15-SP4-SAP-BYOS.x86_64-0.9.3-EC2-Build150400.1.31.raw.xz
-            re.compile(r'''^openqa-SLES
-                              (?P<version>\d+(-SP\d+)?)
-                              (-(?P<type>[^\.]+))?
-                              \.
-                              (?P<arch>[^-]+)
-                              -
-                              (?P<kiwi>\d+\.\d+\.\d+)
-                              -
-                              (?P<flavor>EC2[-\w]*)
-                              -Build(\d+\.)?
-                              (?P<build>\d+\.\d+)
-                              \.raw\.xz
-                              ''', re.RegexFlag.X),
-            # openqa-SLES12-SP4-EC2-HVM-BYOS.x86_64-0.9.2-Build2.56.raw.xz'
-            re.compile(r'''^openqa-SLES
-                              (?P<version>\d+(-SP\d+)?)
-                              -
-                              (?P<flavor>EC2[^\.]+)
-                              \.
-                              (?P<arch>[^-]+)
-                              -
-                              (?P<kiwi>\d+\.\d+\.\d+)
-                              -
-                              Build
-                              (?P<build>\d+\.\d+)
-                              \.raw\.xz
-                              ''', re.RegexFlag.X)
-        ]
-        return self.parse_image_name_helper(img_name, regexes)
-
     def cleanup_all(self):
-        cleanup_ec2_max_snapshot_age_days = PCWConfig.get_feature_property('cleanup', 'ec2-max-snapshot-age-days',
-                                                                           self._namespace)
-        cleanup_ec2_max_volumes_age_days = PCWConfig.get_feature_property('cleanup', 'ec2-max-volumes-age-days',
-                                                                          self._namespace)
-        self.cleanup_images()
-        if cleanup_ec2_max_snapshot_age_days >= 0:
-            self.cleanup_snapshots(cleanup_ec2_max_snapshot_age_days)
-        if cleanup_ec2_max_volumes_age_days >= 0:
-            self.cleanup_volumes(cleanup_ec2_max_volumes_age_days)
+        valid_period_days = PCWConfig.get_feature_property('cleanup', 'ec2-max-age-days', self._namespace)
+
+        if valid_period_days > 0:
+            self.cleanup_images(valid_period_days)
+            self.cleanup_snapshots(valid_period_days)
+            self.cleanup_volumes(valid_period_days)
         if PCWConfig.getBoolean('cleanup/vpc_cleanup', self._namespace):
             self.cleanup_uploader_vpcs()
 
@@ -389,25 +323,13 @@ class EC2(Provider):
                                                                                           region)
                         send_mail('VPC deletion locked by running VMs', body)
 
-    def cleanup_images(self):
+    def cleanup_images(self, valid_period_days):
         for region in self.all_regions:
             response = self.ec2_client(region).describe_images(Owners=['self'])
-            images = list()
             for img in response['Images']:
-                # img is in the format described here:
-                # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_images
-                m = self.parse_image_name(img['Name'])
-                if m:
-                    self.log_dbg("Image {} is candidate for deletion with build {}", img['Name'], m['build'])
-                    images.append(
-                        Image(img['Name'], flavor=m['key'], build=m['build'], date=parse(img['CreationDate']),
-                              img_id=img['ImageId']))
-                else:
-                    self.log_err(" Unable to parse image name '{}'", img['Name'])
-            keep_images = self.get_keeping_image_names(images)
-            for img in [i for i in images if i.name not in keep_images]:
-                self.log_dbg("Delete image '{}' (ami:{})".format(img.name, img.id))
-                if self.dry_run:
-                    self.log_info("Image deletion {} skipped due to dry run mode", img.id)
-                else:
-                    self.ec2_client(region).deregister_image(ImageId=img.id, DryRun=False)
+                if EC2.is_outdated(parse(img['CreationDate']), valid_period_days):
+                    if self.dry_run:
+                        self.log_info("Image deletion {} skipped due to dry run mode", img['ImageId'])
+                    else:
+                        self.log_dbg("Delete image '{}' (ami:{})".format(img['Name'], img['ImageId']))
+                        self.ec2_client(region).deregister_image(ImageId=img['ImageId'], DryRun=False)
--- a/ocw/lib/azure.py
+++ b/ocw/lib/azure.py
@@ -1,4 +1,4 @@
-from .provider import Provider, Image
+from .provider import Provider
 from webui.settings import PCWConfig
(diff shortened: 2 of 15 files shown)


mimi1vx:
I'm all pro for drop of unneded code :D

ocw/lib/gce.py
```diff
@@ -154,50 +98,26 @@ def cleanup_all(self):
...
-                        image["name"],
-                        m["build"],
-                    )
-                else:
-                    self.log_err("Unable to parse image name '{}'", image["name"])
+                        response = request.execute()
+                        if "error" in response:
+                            for e in response["error"]["errors"]:
+                                self.log_err(e["message"])
+                        if "warnings" in response:
```
ricardobranco777, line 115:
```suggestion
                        elif "warnings" in response:
```

ocw/lib/azure.py
```diff
@@ -100,126 +100,44 @@ def list_disks_by_resource_group(self, resource_group):
...
+                            self.container_client(c.name).delete_blob(blob.name, delete_snapshots="include")
 
-    def cleanup_images_from_rg(self, keep_images):
+    def cleanup_images_from_rg(self):
         for item in self.list_images_by_resource_group(self.__resource_group):
-            m = self.parse_image_name(item.name)
-            if m:
-                self.log_dbg('Image {} is candidate for deletion with build {} ', item.name, m['build'])
-                if item.name not in keep_images:
+            if (self.older_than(item.changed_time)):
```
ricardobranco777, line 126:
```suggestion
            if self.older_than(item.changed_time):
```

ocw/lib/azure.py
```diff
@@ -100,126 +100,44 @@ def list_disks_by_resource_group(self, resource_group):
...
-                    self.log_info("Delete blob '{}'", img.name)
-                    if self.dry_run:
-                        self.log_info("Deletion of blob image {} skipped due to dry run mode", img.name)
-                    else:
-                        container_client.delete_blob(img.name, delete_snapshots="include")
+            if 'pcw_ignore' not in c['metadata']:
+                self.log_dbg('Found container {}', c.name)
+                container_blobs = self.container_client(c.name).list_blobs()
+                for blob in container_blobs:
+                    if (self.older_than(blob.last_modified)):
```
ricardobranco777, line 117:
```suggestion
                    if self.older_than(blob.last_modified):
```

ricardobranco777, line 117:
```suggestion
                    if self.older_than(blob.last_modified):
```

ricardobranco777 (approved):
Small suggestions. Otherwise LGTM

ocw/lib/gce.py
asmorodskyi: my understanding that there could be errors AND warnings at the same time so not a good idea

ocw/lib/azure.py
```diff
@@ -100,126 +99,44 @@ def list_disks_by_resource_group(self, resource_group):
...
+                            self.container_client(c.name).delete_blob(blob.name, delete_snapshots="include")
 
-    def cleanup_images_from_rg(self, keep_images):
+    def cleanup_images_from_rg(self):
         for item in self.list_images_by_resource_group(self.__resource_group):
-            m = self.parse_image_name(item.name)
-            if m:
-                self.log_dbg('Image {} is candidate for deletion with build {} ', item.name, m['build'])
-                if item.name not in keep_images:
+            if (self.older_than_max_age_hours(item.changed_time)):
```
ricardobranco777, line 125:
```suggestion
            if self.older_than_max_age_hours(item.changed_time):
```

ocw/lib/azure.py
```diff
@@ -100,126 +99,44 @@ def list_disks_by_resource_group(self, resource_group):
...
+    def cleanup_disks_from_rg(self):
         for item in self.list_disks_by_resource_group(self.__resource_group):
-            m = self.parse_image_name(item.name)
-            if m:
-                self.log_dbg('Disk {} is candidate for deletion with build {} ', item.name, m['build'])
-
-                if item.name not in keep_images:
-                    if self.compute_mgmt_client().disks.get(self.__resource_group, item.name).managed_by:
-                        self.log_warn("Disk is in use - unable delete {}", item.name)
+            if (self.older_than_max_age_hours(item.changed_time)):
```
ricardobranco777, line 134:
```suggestion
            if self.older_than_max_age_hours(item.changed_time):
```

ricardobranco777 (commented):
Some small suggestions.  These could be easily catched with flake8 or pylint.

ocw/lib/EC2.py
```diff
@@ -209,66 +196,13 @@ def delete_all_clusters(self):
...
-        cleanup_ec2_max_snapshot_age_days = PCWConfig.get_feature_property('cleanup', 'ec2-max-snapshot-age-days',
-                                                                           self._namespace)
-        cleanup_ec2_max_volumes_age_days = PCWConfig.get_feature_property('cleanup', 'ec2-max-volumes-age-days',
-                                                                          self._namespace)
-        self.cleanup_images()
-        if cleanup_ec2_max_snapshot_age_days >= 0:
-            self.cleanup_snapshots(cleanup_ec2_max_snapshot_age_days)
-        if cleanup_ec2_max_volumes_age_days >= 0:
-            self.cleanup_volumes(cleanup_ec2_max_volumes_age_days)
+        cleanup_ec2_max_age_days = PCWConfig.get_feature_property('cleanup', 'ec2-max-age-days', self._namespace)
```
cfconrad, line 200: why do we need the extra variable for ec2 here?

templates/pcw.ini
```diff
@@ -17,26 +17,41 @@ smtp-port = 25
...
 namespaces = qac
 
 # Values specified under "cleanup.namespace.{namespace}" have precedence over same values in [cleanup] for this certain namespace
 [cleanup.namespace.qac]
-# EC2 snapshots younger than this amount of days will be ignored
-ec2-max-snapshot-age-days = 2
+# Max age of an image file ( used in EC2 only )
+ec2-max-age-days = 2
 # EC2 volumes younger than this amount of days will be ignored
 ec2-max-volumes-age-days = 2
```
cfconrad, line 43: I think this value isn't there as well or?
```suggestion

```

ocw/lib/EC2.py
asmorodskyi: historical reasons , I think we will keep this for next commits ...

templates/pcw.ini
asmorodskyi: yes , you right , removing

ocw/lib/EC2.py
```diff
@@ -89,45 +88,33 @@ def all_clusters(self):
...
-            regexes = [
-                re.compile(r'''^OpenQA upload image$'''),
-                re.compile(r'''^Created by CreateImage\([\w-]+\) for ami-\w+ from vol-\w+$''')
-            ]
-            for regex in regexes:
-                m = re.match(regex, snapshot['Description'].strip())
-                if m:
-                    return True
-        return False
+    def needs_to_delete_by_age(creation_time, cleanup_ec2_max_age_days) -> bool:
```
grisu48, line 91:
```suggestion
    def is_outdated(creation_time, cleanup_ec2_max_age_days) -> bool:
```

Perhaps make it clearer, that this is a comparison method

ocw/lib/EC2.py
```diff
@@ -209,66 +196,13 @@ def delete_all_clusters(self):
...
-        cleanup_ec2_max_volumes_age_days = PCWConfig.get_feature_property('cleanup', 'ec2-max-volumes-age-days',
-                                                                          self._namespace)
-        self.cleanup_images()
-        if cleanup_ec2_max_snapshot_age_days >= 0:
-            self.cleanup_snapshots(cleanup_ec2_max_snapshot_age_days)
-        if cleanup_ec2_max_volumes_age_days >= 0:
-            self.cleanup_volumes(cleanup_ec2_max_volumes_age_days)
+        cleanup_ec2_max_age_days = PCWConfig.get_feature_property('cleanup', 'ec2-max-age-days', self._namespace)
+
+        if cleanup_ec2_max_age_days >= 0:
```
grisu48, line 202:
```suggestion
        if cleanup_ec2_max_age_days > 0:
```

I don't think that setting this value to `0` makes any sense except for disabling this check or am I wrong? Is `0` a valid setting?

ocw/lib/azure.py
```diff
@@ -100,126 +99,44 @@ def list_disks_by_resource_group(self, resource_group):
...
+    def cleanup_disks_from_rg(self):
         for item in self.list_disks_by_resource_group(self.__resource_group):
-            m = self.parse_image_name(item.name)
-            if m:
-                self.log_dbg('Disk {} is candidate for deletion with build {} ', item.name, m['build'])
-
-                if item.name not in keep_images:
-                    if self.compute_mgmt_client().disks.get(self.__resource_group, item.name).managed_by:
-                        self.log_warn("Disk is in use - unable delete {}", item.name)
+            if self.older_than_max_age_hours(item.changed_time):
```
grisu48, line 134:
```suggestion
            if self.is_outdated(item.changed_time):
```

I think that name would be more descriptive, WDYT?

ocw/lib/gce.py
```diff
@@ -89,115 +90,33 @@ def delete_instance(self, instance_id, zone):
...
-                # name:sles12-sp5-gce-x8664-0-9-1-byos-build1-56
-                m = self.parse_image_name(image["name"])
-                if m:
-                    images.append(
-                        Image(
-                            image["name"],
-                            flavor=m["key"],
-                            build=m["build"],
-                            date=parse(image["creationTimestamp"]),
+                if self.older_than_max_age_hours(parse(image["creationTimestamp"]).astimezone(timezone.utc)):
```
grisu48, line 100:
```suggestion
                if self.is_outdated(parse(image["creationTimestamp"]).astimezone(timezone.utc)):
```
More descriptive (and shorter!) method name suggestion

ocw/lib/provider.py
```diff
@@ -31,48 +29,17 @@ def getData(self, name=None):
...
-        keep_images = list()
-        for img_list in [images_by_flavor[x] for x in sorted(images_by_flavor)]:
-            img_list.sort(key=lambda x: LooseVersion(x.build), reverse=True)
-            for i in range(0, len(img_list)):
-                img = img_list[i]
-                if (not self.needs_to_delete_image(i, img.date)):
-                    keep_images.append(img.name)
-
-        return keep_images
+    def older_than_max_age_hours(self, age):
```
grisu48, line 32:
```suggestion
    def is_outdated(self, age):
```

ocw/lib/provider.py
```diff
@@ -31,48 +29,17 @@ def getData(self, name=None):
...
+        """
+            older_than_max_age_hours - calculates if certain resource bypass maximum allowed TTL
+            maximum allowed TTL is controled by cleanup/max-age-hours pcw.ini config param
+            :param age: usually creation time of resource or any other timestamp which may be used to identify
+                age of the resource
+            :return: True if resource is already too old , false otherwise
+        """
+        delta_in_hours = PCWConfig.get_feature_property('cleanup', 'max-age-hours', self._namespace)
+        max_allowed_age = datetime.now(timezone.utc) - timedelta(hours=delta_in_hours)
+        return max_allowed_age > age
```
grisu48, line 42: The name of the method is `older_than_max_age_hours` but this comparison does exactly the opposite :thinking:

templates/pcw.ini
```diff
@@ -17,26 +17,39 @@ smtp-port = 25
...
 # list of recipients which will be notified in case of issues in certain namespace
 to = <your_email@suse.com>
 
 [cleanup]
-# Specify how many images per flavor get kept
-max-images-per-flavor = 2
-# Max age of an image file
-max-images-age-hours = 24
+# Max age of data storage resources ( used in EC2 only )
+ec2-max-age-days = 1
```
grisu48, line 31: Perhaps not for this PR, but why not define ONE value in `dd:hh:mm` instead of scattering the allowed timestamp over multiple values?

e.g.

```suggestion
ec2-max-age = 1d
```

templates/pcw.ini
```diff
@@ -17,26 +17,39 @@ smtp-port = 25
...
+ec2-max-age-days = 2
+# Azure resource group which PCW will scan for data storage resources needs to be deleted
 azure-storage-resourcegroup = openqa-upload
+# AccountName used for creation of BlobServiceClient
 azure-storage-account-name = openqa
 # When set to true EC2 VPC cleanup will be enabled
 vpc_cleanup = true
+
+[updaterun]
+# if openqa_ttl tag is not defined this TTL will be set to the instance
```
grisu48, line 50: Please define, what unit this ttl is in (hours? days? bananas?)

tests/test_ec2.py
```diff
@@ -1,13 +1,16 @@
 from ocw.lib.EC2 import EC2, Provider
 from webui.settings import PCWConfig
 from tests.generators import mock_get_feature_property
-from tests.generators import min_image_age_hours, max_image_age_hours, ec2_max_volumes_age_days, \
-    ec2_max_snapshot_age_days
+from tests.generators import ec2_max_age_days
+from faker import Faker
```
grisu48, line 5: Why use `faker`, i.e. another dependency instead of the build-in [unittest.mock](https://docs.python.org/3/library/unittest.mock.html) ?

tests/test_gce.py
```diff
@@ -60,28 +32,26 @@ def delete(self, *args, **kwargs):
 
 
 def test_cleanup_all(monkeypatch):
-    newer_then_min_age = datetime.now(timezone.utc).isoformat()
-    older_then_min_age = (datetime.now(timezone.utc) - timedelta(hours=min_image_age_hours+1)).isoformat()
-    older_then_max_age = (datetime.now(timezone.utc) - timedelta(hours=max_image_age_hours+1)).isoformat()
+    older_then_max_age = (datetime.now(timezone.utc) - timedelta(hours=max_age_hours+1)).strftime("%m/%d/%Y, %H:%M:%S")
```
grisu48, line 35:
```suggestion
    older_than_max_age = (datetime.now(timezone.utc) - timedelta(hours=max_age_hours+1)).strftime("%m/%d/%Y, %H:%M:%S")
```

grisu48 (commented):
Overall: I like PRs that delete more code than adding :+1: :slightly_smiling_face: 

I left some comment and some questions though but in general this looks like a good improvement to me.

ocw/lib/EC2.py
```diff
@@ -89,45 +88,33 @@ def all_clusters(self):
...
 
-    def cleanup_volumes(self, cleanup_ec2_max_volumes_age_days):
-        delete_older_than = date.today() - timedelta(days=cleanup_ec2_max_volumes_age_days)
+    def cleanup_volumes(self, cleanup_ec2_max_age_days):
         for region in self.all_regions:
             response = self.ec2_client(region).describe_volumes()
             for volume in response['Volumes']:
-                if datetime.date(volume['CreateTime']) < delete_older_than:
+                if EC2.needs_to_delete_by_age(volume['CreateTime'], cleanup_ec2_max_age_days):
                     if self.volume_protected(volume):
```
grisu48, line 118: I guess this can be deleted.

ocw/lib/azure.py
```diff
@@ -100,126 +99,44 @@ def list_disks_by_resource_group(self, resource_group):
...
-        keep_images = self.get_keeping_image_names()
-        self.cleanup_sle_images_container(keep_images)
-        self.cleanup_disks_from_rg(keep_images)
-        self.cleanup_images_from_rg(keep_images)
-        for i in keep_images:
-            self.log_info("Keep image {} ", i)
-
-    def cleanup_bootdiagnostics(self):
-        containers = self.bs_client().list_containers()
+        self.cleanup_blob_containers()
```
grisu48, line 105: I think blobs need to be deleted at last.

ocw/lib/azure.py
```diff
@@ -100,126 +99,44 @@ def list_disks_by_resource_group(self, resource_group):
...
-
-                if img.name not in keep_images:
-                    self.log_info("Delete blob '{}'", img.name)
-                    if self.dry_run:
-                        self.log_info("Deletion of blob image {} skipped due to dry run mode", img.name)
-                    else:
-                        container_client.delete_blob(img.name, delete_snapshots="include")
+            if 'pcw_ignore' not in c['metadata']:
+                self.log_dbg('Found container {}', c.name)
+                container_blobs = self.container_client(c.name).list_blobs()
```
grisu48, line 114: Before we only looked for blobs in the `sle-images` container. Please also limit the behaviour here to only look for this container.

ocw/lib/provider.py
grisu48: There is a bit of confusion in this method - `age` is not an age (i.e. a duration) but a `timestamp`, i.e. a certain point in time. Suggestion:

```
def is_outdated(self, timestamp):
    delta = PCWConfig.get_feature_property('cleanup', 'max-age-hours', self._namespace)       # maximum allowed delta
    delta = timedelta(delta)
    now = datetime.now(timezone.utc)
    # Resource is outdated, if the given timestamp plus allowed delta exceeds the current timestamp
    return (timestamp + delta) > now
```

tests/test_ec2.py
grisu48: Probably beyond the scope of this already big PR.

ocw/lib/EC2.py
asmorodskyi: done

ocw/lib/EC2.py
asmorodskyi: after thinking more about that , I decided to keep it . It will basically drop ability to protect volumes which are not attached to running VM . I renamed tag to be "pcw_ignore" to align this with all other places where we have similar logic . If you want to insist on deletion let's have separate chat about this

ocw/lib/EC2.py
asmorodskyi: fixed

ocw/lib/EC2.py
```diff
@@ -209,66 +196,13 @@ def delete_all_clusters(self):
...
-        if cleanup_ec2_max_snapshot_age_days >= 0:
-            self.cleanup_snapshots(cleanup_ec2_max_snapshot_age_days)
-        if cleanup_ec2_max_volumes_age_days >= 0:
-            self.cleanup_volumes(cleanup_ec2_max_volumes_age_days)
+        cleanup_ec2_max_age_days = PCWConfig.get_feature_property('cleanup', 'ec2-max-age-days', self._namespace)
+
+        if cleanup_ec2_max_age_days >= 0:
+            self.cleanup_images(cleanup_ec2_max_age_days)
+            self.cleanup_volumes(cleanup_ec2_max_age_days)
+            self.cleanup_snapshots(cleanup_ec2_max_age_days)
```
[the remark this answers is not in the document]
asmorodskyi: done

ocw/lib/azure.py
asmorodskyi: done

ocw/lib/azure.py
asmorodskyi: done ( see  container_valid_for_cleanup )

ocw/lib/provider.py
asmorodskyi: the code which you suggesting will not give expected behavior . I can show you in example . 
Let's say `max-age-hours=7`  and if I call it with `is_outdated("01-01-2022 21-00")` it will gives me false . your method may be called `is_still_valid` because it will return true only for timestamps which didn't cross max-age-hours

ocw/lib/provider.py
asmorodskyi: done

ocw/lib/gce.py
asmorodskyi: done

ocw/lib/azure.py
asmorodskyi: done

templates/pcw.ini
asmorodskyi: done

tests/test_gce.py
asmorodskyi: done

ocw/lib/EC2.py
grisu48: "It will basically drop ability to protect volumes which are not attached to running VM" - I have no idea what you are trying to say with this sentence. What does `It` mean? What does `protect` means in this context?

ocw/lib/provider.py
grisu48: True, the last line should be `return now > (timestamp > delta)`

templates/pcw.ini
grisu48: https://github.com/SUSE/pcw/issues/174

ocw/lib/EC2.py
asmorodskyi: if you need to some volume not to be deleted forever , currently you can label it with "DO_NOT_DELETE" and than PCW will exclude it from its queries , so volume will be *protected* from PCW which otherwise will delete it after some time . If we simply drop this function together with it we will drop mechanism allowing to protect volumes from PCW

grisu48: Now I understand and now it also makes sense. Thanks, yes let's keep it! Marking as resolved.

PR #172, opened 2022-11-29
Add flake8 linter

pdostal:
Added as suggested in #172


asmorodskyi:
I think you need to make it green in same PR https://github.com/SUSE/pcw/actions/runs/3572419973/jobs/6005285498

grisu48 (approved):

PR #175, opened 2022-12-02
K8s cleanup

ilausuch:
https://progress.opensuse.org/issues/120660

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ htmlcov/*
 # Configuration and program data
 /db
 pcw.ini
+
+# Development
+run/
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ COPY requirements.txt /pcw/
 # * Install system requirements
 # * Install pip requirements
 # * Empty system cache to conserve some space
-RUN zypper -n in python310-devel gcc libffi-devel && pip3.10 install -r /pcw/requirements.txt && rm -rf /var/cache
+RUN zypper -n in python310-devel gcc libffi-devel aws-cli && pip3.10 install -r /pcw/requirements.txt && rm -rf /var/cache
 
 # Copy program files only
 COPY ocw  /pcw/ocw/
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -11,7 +11,7 @@ COPY requirements.txt /pcw/
 # * Install pip requirements
 # * Empty system cache to conserve some space
 RUN zypper -n in python310-devel gcc libffi-devel && pip3.10 install -r /pcw/requirements.txt && rm -rf /var/cache
-
+RUN zypper in -y kubernetes1.23-client
```
asmorodskyi, line 14: just include this in list above , this will decrease build time

ocw/lib/EC2.py
```diff
@@ -411,3 +412,69 @@ def cleanup_images(self):
...
+
+                    config_file = f"/tmp/k8s_{region}_{cluster_name}.yaml"
+
+                    # Write in YAML.
+                    config_text=yaml.dump(cluster_config, default_flow_style=False)
+                    open(config_file, "w").write(config_text)
+
+                    self.log_info(f"Created config file {config_file}")
+
+                    resp = subprocess.call(["kubectl", "--kubeconfig=%s" % config_file, "get", "nodes"])
```
asmorodskyi, line 475: maybe you look for some python lib for interacting with k8s ?

asmorodskyi: something like this https://github.com/kubernetes-client/python ? But please do better research it is just first URL from google

Dockerfile
pdostal: Build time and number of layers :smile:

.gitignore
```diff
@@ -17,3 +17,6 @@ htmlcov/*
 # Configuration and program data
 /db
 pcw.ini
+
+# Development
+run/
```
asmorodskyi, line 22: we don't need this . normally you running things inside container

PR #181, opened 2023-01-06
Bring order into log messages

asmorodskyi:


ocw/lib/EC2.py
```diff
@@ -134,8 +138,10 @@ def cleanup_snapshots(self, valid_period_days):
                             raise ex
 
     def cleanup_volumes(self, valid_period_days):
+        self.log_dbg("Call cleanup volumes")
```
pdostal, line 141:
```suggestion
        self.log_dbg("Call cleanup_volumes")
```

ocw/lib/db.py
```diff
@@ -224,8 +224,8 @@ def auto_delete_instances():
                          age__gte=F('ttl')).exclude(csp_info__icontains='pcw_ignore')
         email_text = set()
         for i in obj:
-            logger.info("[%s] TTL expire for instance %s:%s %s", i.vault_namespace,
-                        i.provider, i.instance_id, i.all_time_fields())
+            logger.debug("[%s] TTL expire for instance %s:%s %s", i.vault_namespace,
```
pdostal, line 227: I would keep this log as `info`.

ocw/lib/db.py
```diff
@@ -161,7 +161,7 @@ def _update_provider(name, vault_namespace):
         for region in EC2(vault_namespace).all_regions:
             instances_csp = EC2(vault_namespace).list_instances(region=region)
             instances += [ec2_to_local_instance(i, vault_namespace, region) for i in instances_csp]
-            logger.info("Got %d instances from EC2 in region %s", len(instances), region)
+        logger.info("Got %d instances from EC2", len(instances))
```
pdostal, line 164: Why? I would keep it more verbose.

pdostal (approved):
I've few nits only.

ocw/lib/db.py
asmorodskyi: 1. because it is a fake , if you look closely you will see that we sum together all entities in one value but in log you see like each region has that amount of entities 
2.  from 2 years personal experience reading logs I don't need such details per region , total amount of entities is more than enough

ocw/lib/db.py
asmorodskyi: one line below there is delete_instance call which will gives you notification about deletion of entity anyway , now we getting 2 lines per one deletion it is too much

PR #188, opened 2023-01-24
Add comment for SECRET_KEY setting

ricardobranco777:
Add comment for SECRET_KEY setting

### DIFF
--- a/webui/settings.py
+++ b/webui/settings.py
@@ -24,7 +24,7 @@ BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/2.1/howto/deployment/checklist/
 
-# SECURITY WARNING: keep the secret key used in production secret!
+# See https://docs.djangoproject.com/en/dev/ref/settings/#secret-key
 SECRET_KEY = os.getenv('SECRET_KEY')
 
 # SECURITY WARNING: don't run with debug turned on in production!


webui/settings.py
```diff
@@ -25,7 +25,10 @@
 # See https://docs.djangoproject.com/en/2.1/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
```
asmorodskyi, line 27: I think with code and comments above we can remove this warning

webui/settings.py
```diff
@@ -25,7 +25,10 @@
...
-SECRET_KEY = os.getenv('SECRET_KEY')
+# DO NOT REMOVE if you're setting the password with Ansible, so Django creates the auth_user table;
+#   otherwise use the hashed password.
+# See https://docs.djangoproject.com/en/dev/ref/settings/#secret-key
+SECRET_KEY = os.getenv('SECRET_KEY', 'DO NOT REMOVE')
```
asmorodskyi, line 31: I am not sure it is good idea , currently if variable is not defined anywhere you will get error that you must define it so you will clearly know what is going wrong . with this default value you will calm down django and your env will work even without variable defined but you may get troubles to login because you don't know the password because you haven't explicitly set it anywhere

webui/settings.py
ricardobranco777: Agree

webui/settings.py
ricardobranco777: > I am not sure it is good idea , currently if variable is not defined anywhere you will get error that you must define it so you will clearly know what is going wrong . with this default value you will calm down django and your env will work even without variable defined but you may get troubles to login because you don't know the password because you haven't explicitly set it anywhere

I agree.  Will update the MR for the Ansible deployment code to set some dummy value.

PR #209, opened 2023-02-16
Multiple improvements

ricardobranco777:
This PR:
- Adds `createuser` to `container-startup` as we don't need the Django superuser anymore.
- Mitigates the `500 Server Error` we're getting because of the use of SQLite.  A full fix requires the migration to PostgreSQL.
- Fixes the use of `SECRET_KEY` which is supposed to be randomly generated according to the [documentation](https://docs.djangoproject.com/en/dev/ref/settings/#secret-key).

Related MR:
https://gitlab.suse.de/qac/publiccloud-qa-suse-de/-/merge_requests/66


container-startup
```diff
@@ -5,15 +5,15 @@
...
+usage() {
 	echo "pcw startup helper"
 	echo "Usage: $0 COMMAND"
 	echo "COMMANDS:"
 	echo "   help                    Print help message"
 	echo "   check                   Run system health checks"
 	echo "   run                     Standart publiccloud-watcher run procedure"
 	echo "   migrate                 Run database migration"
-	echo "   createsuperuser         Create superuser. Requires --email and --username argument"
+	echo "   createuser USER PASS    Create user"
```
asmorodskyi, line 16: if idea to have only single user why do we need to have special command to add users ?

container-startup
```diff
@@ -5,15 +5,15 @@
 CMD=$1
 shift
 
-function usage() {
+usage() {
```
asmorodskyi, line 8: can you elaborate why it is necessary ? ( generally I don't mind to remove it just want to know what is difference )

container-startup
ricardobranco777: Because otherwise we would have to run `python3 manage.py shell -c "from django.contrib.auth.models import User; user = User.objects.create_user('USERNAME', password='PASSWORD'); user.save()"` which is ugly.

container-startup
ricardobranco777: Just to make the `shellcheck` tool happy:

```
In container-startup line 8:
function usage() {
^-- SC2112 (warning): 'function' keyword is non-standard. Delete it.
```

webui/settings.py
```diff
@@ -11,21 +12,21 @@
...
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
 
 # Quick-start development settings - unsuitable for production
-# See https://docs.djangoproject.com/en/2.1/howto/deployment/checklist/
+# See https://docs.djangoproject.com/en/4.0/howto/deployment/checklist/
 
 # See https://docs.djangoproject.com/en/dev/ref/settings/#secret-key
-SECRET_KEY = os.getenv('SECRET_KEY')
+SECRET_KEY = utils.get_random_secret_key()
```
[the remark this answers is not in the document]
ricardobranco777: `SECRET_KEY` is _not_ a password.  The password is set by the `createuser` function.

The [documentation](https://docs.djangoproject.com/en/dev/ref/settings/#secret-key) states that:
> A secret key for a particular Django installation. This is used to provide cryptographic signing, and should be set to a unique, unpredictable value.

container-startup
asmorodskyi: ok , make sense :+1:

container-startup
asmorodskyi: no I mean we already have "createsuperuser" and we don't want to have other users at all so why do we need to have "createuser" at all ? who and when will call it ?

webui/settings.py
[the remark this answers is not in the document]
asmorodskyi: ah ok I see ! I always thought that this is our password :D

container-startup
ricardobranco777: > no I mean we already have "createsuperuser" and we don't want to have other users at all so why do we need to have "createuser" at all ? who and when will call it ?

The problem with `createsuperuser` is that it creates a special Django superuser which we don't need at all.  A normal user as created by `createuser` is enough.  The Django superuser is needed only for the Django admin interface that we don't use.

container-startup
```diff
@@ -47,8 +47,13 @@ case "$CMD" in
 	"migrate")
 		python3 manage.py migrate
 		;;
-	"createsuperuser")
-		python3 manage.py createsuperuser $@
+	"createuser")
```
asmorodskyi, line 50: this will requires PR into https://gitlab.suse.de/qac/publiccloud-qa-suse-de/

PR #211, opened 2023-02-20
Add dedicated container to cleanup k8s jobs

asmorodskyi:


.github/workflows/container.yml
```diff
@@ -9,7 +9,8 @@ name: Create and publish container
 
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME: ${{ github.repository }}
+  PCW_IMAGE_NAME: ${{ github.repository }}_main
+  K8S_GCE_IMAGE_NAME : ${{ github.repository }}_gce_clean
```
ilausuch, line 13: I think should be generic for all the providers, as far as I know Dockerfile_k8s will be a generic for all of them

PR #213, opened 2023-02-21
Move EKS jobs cleanup to cleanup_k8s.py script

ilausuch:
https://progress.opensuse.org/issues/124664


ocw/lib/eks.py
```diff
@@ -0,0 +1,98 @@
...
+                                f"--kubeconfig {kubeconfig}")
+            if res.returncode != 0:
+                raise Exception(f"Cannot get the kubeconfig for the cluster {cluster_name} on region {region}")
+
+            kubernetes.config.load_kube_config(config_file=kubeconfig)
+            self.__kubectl_client[region_cluster] = kubernetes.client.BatchV1Api()
+
+        return self.__kubectl_client[region_cluster]
+
+    def all_clusters(self) -> dict:
```
asmorodskyi, line 73: well this piece is not directly related to what you are moving ... 
Generally I am fine with this move but then you also need to do something with `list_clusters()` function  currently you just broke it

ilausuch: yes it's correct

ilausuch: done

ilausuch:
I am working on the test for eks

ocw/lib/eks.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def kubectl_client(self, region: str, cluster_name: str):
+        region_cluster = f"{region}/{cluster_name}"
+
+        if region_cluster not in self.__kubectl_client:
+            kubeconfig = f"~/.kube/eks_config_{region}_{cluster_name}"
+            kubeconfig = os.path.expanduser(kubeconfig)
+
+            res = self.cmd_exec(f"aws eks update-kubeconfig --region {region} --name {cluster_name} " +
+                                f"--kubeconfig {kubeconfig}")
```
ricardobranco777, line 74:
```suggestion
            res = self.cmd_exec(f"aws eks update-kubeconfig --region {region} --name {cluster_name} --kubeconfig {kubeconfig}")
```

ricardobranco777 (approved):
Just a small suggestion. Otherwise LGTM

ocw/lib/eks.py
```diff
@@ -0,0 +1,90 @@
...
+        if self.__cluster_regions is None:
+            regions_query = self.cmd_exec(
+                f"aws ec2 describe-regions --query 'Regions[].RegionName' --output json --region {EKS.default_region}")
+            self.__cluster_regions = json.loads(regions_query.stdout)
+        return self.__cluster_regions
+
+    def create_credentials_file(self) -> None:
+        creds_file = f"{self.aws_dir()}/credentials"
+
+        if not os.path.exists(creds_file):
```
asmorodskyi, line 44: and do you need to check if file exists to decide about dir creation ? 

what will happen if we will keep it like this : 
```
if not os.path.exists(self.aws_dir()):
                os.mkdir(self.aws_dir())
```

cleanup_k8s.py
```diff
@@ -12,6 +13,8 @@ def main():
         try:
             if ProviderChoice.GCE in providers:
```
asmorodskyi, line 14: this won't work right anymore , you need loop on providers . if you will just add another "if" here . only first one will be processed than

ocw/lib/eks.py
ilausuch: Yes, but my initial intention is to create the credentials if not exists, so was an indentation problem. In next commit will be fixed

cleanup_k8s.py
ilausuch: yes, you are rigth

asmorodskyi: ah , you right . sorry

ocw/lib/eks.py
asmorodskyi: ok now it make sense :) . python :)

ocw/management/commands/cleanupk8s.py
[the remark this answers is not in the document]
ilausuch: Because is is done outside the regular pcw process by cleanup_k8s.py. Also needs an special container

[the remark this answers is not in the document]
asmorodskyi: ah right and there is no django in k8s cleanup . you right

PR #217, opened 2023-02-22
Run containers as non-root & reduce the size of pcw image

ricardobranco777:
This PR:
- Makes it possible to run the containers as non-root
- Adds `--no-cache-dir` to `pip install` to reduce the size by 53M

Fixes: https://github.com/SUSE/pcw/issues/206

Related MR:
https://gitlab.suse.de/qac/publiccloud-qa-suse-de/-/merge_requests/70

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,5 +28,8 @@ VOLUME /pcw/db
 
 EXPOSE 8000/tcp
 
+RUN useradd --no-create-home --uid 1777 --user-group --shell /bin/false pcw && chown -R pcw:pcw /pcw
+USER pcw
+
 # Once we are certain that this runs nicely, replace this with ENTRYPOINT.
 CMD ["/pcw/container-startup", "run"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ COPY requirements.txt /pcw/
 # * Install system requirements
 # * Install pip requirements
 # * Empty system cache to conserve some space
-RUN zypper -n in python310-devel gcc libffi-devel aws-cli && pip3.10 install -r /pcw/requirements.txt && rm -rf /var/cache
+RUN zypper -n in python310-devel gcc libffi-devel aws-cli && pip install --no-cache-dir -r /pcw/requirements.txt && rm -rf /var/cache
 
 # Copy program files only
 COPY ocw  /pcw/ocw/
--- a/Dockerfile_dev
+++ b/Dockerfile_dev
@@ -14,4 +14,7 @@ RUN zypper -n in python310-devel gcc libffi-devel aws-cli && pip install --no-ca
 
 WORKDIR /pcw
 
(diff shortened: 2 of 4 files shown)


pdostal:
Is `--no-cache-dir` necessary when we have `&& rm -rf /var/cache` at the end of that command?

Dockerfile
```diff
@@ -28,5 +28,8 @@ VOLUME /pcw/db
 
 EXPOSE 8000/tcp
 
+RUN useradd --no-create-home --uid 1777 --user-group --shell /bin/false pcw && chown -R pcw:pcw /pcw
+USER pcw
```
pdostal, line 32: This is unrelated to current pull request. I'm all for it but I think this should be separate.

pdostal (approved):

Dockerfile_dev
```diff
@@ -10,7 +10,7 @@ COPY requirements.txt requirements_test.txt requirements_k8s.txt /tmp/
 # * Install system requirements
 # * Install pip requirements
 # * Empty system cache to conserve some space
-RUN zypper -n in python310-devel gcc libffi-devel aws-cli && pip3.10 install -r /tmp/requirements_test.txt && rm -rf /var/cache
+RUN zypper -n in python310-devel gcc libffi-devel aws-cli && pip install --no-cache-dir -r /tmp/requirements_test.txt && rm -rf /var/cache
```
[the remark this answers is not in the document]
ricardobranco777: > what about _dev container ? I think it should follow same non-root approach so we will catch any user-related problems earlier but I am not sure that it is right thing to use `pcw` user which certainly not exists anywhere . maybe for _dev container we need to build it with user who init `podman build ` ? I think it is possible to propagate via Makefile ... WDYT ?

The related MR (that was already merged by @pdostal takes care of this). I tried this PR & MR it and it works.

Will add it to _dev as well.

Dockerfile
```diff
@@ -10,7 +10,7 @@ COPY requirements.txt /pcw/
 # * Install system requirements
 # * Install pip requirements
 # * Empty system cache to conserve some space
-RUN zypper -n in python310-devel gcc libffi-devel aws-cli && pip3.10 install -r /pcw/requirements.txt && rm -rf /var/cache
+RUN zypper -n in python310-devel gcc libffi-devel aws-cli && pip install --no-cache-dir -r /pcw/requirements.txt && rm -rf /var/cache
```
asmorodskyi, line 13: if you want to introduce user I think you also need to solve this 
```
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
```

asmorodskyi: but instead of virtualenv I would simply run pip AFTER switching to pcw user , don't think it make sense to create virtualenv inside container

asmorodskyi: ok forget it

PR #230, opened 2023-03-22
Remove appscheduler , we will use outside cron scheduling


asmorodskyi:
Remove appscheduler , we will use outside cron scheduling


pdostal (approved):
Not `cron`, we use `systemd-timer` 🎉

asmorodskyi:
it won't work . too many things will get broken

PR #238, opened 2023-04-03
Another attempt to make VPC cleanup working

asmorodskyi:
Another attempt to make VPC cleanup working

1. Add missing check for pcw_ignore flag !
2. re-order deletion of dependent entities
3. remove elastic IP deletion ( looks like it is not needed )

### DIFF
--- a/ocw/lib/EC2.py
+++ b/ocw/lib/EC2.py
@@ -1,6 +1,6 @@
 import traceback
 import time
-from datetime import date, datetime, timedelta, timezone
+from datetime import date, datetime, timedelta
 from typing import Dict
 import boto3
 from botocore.exceptions import ClientError
@@ -147,27 +147,16 @@ class EC2(Provider):
         if PCWConfig.getBoolean('cleanup/vpc_cleanup', self._namespace):
             self.cleanup_vpcs()
 
-    def delete_elastic_ips(self, region):
-        self.log_info('Deleting elastic IPs in {}'.format(region))
-        response = self.ec2_client(region).describe_addresses()
-        for addr in response['Addresses']:
-            if 'AssociationId' in addr:
-                self.log_info('Disassosiate IP with AssociationId:{}'.format(addr['AssociationId']))
-                self.ec2_client(region).disassociate_address(AssociationId=addr['AssociationId'])
-            self.log_info('Release IP with AllocationId:{}'.format(addr['AllocationId']))
-            self.ec2_client(region).release_address(AllocationId=addr['AllocationId'])
-
-    def delete_vpc(self, region: str, vpc, vpc_id: str) -> None:
+    def delete_vpc(self, region: str, vpc, vpc_id: str):
         try:
             self.log_info('{} has no associated instances. Initializing cleanup of it', vpc)
-            self.delete_elastic_ips(region)
-            self.delete_internet_gw(vpc)
-            self.delete_routing_tables(vpc)
-            self.delete_vpc_endpoints(region, vpc_id)
+            self.delete_routing_tables(region, vpc_id)
             self.delete_security_groups(vpc)
-            self.delete_vpc_peering_connections(region, vpc_id)
             self.delete_network_acls(vpc)
             self.delete_vpc_subnets(vpc)
+            self.delete_internet_gw(vpc)
+            self.delete_vpc_endpoints(region, vpc_id)
+            self.delete_vpc_peering_connections(region, vpc_id)
             if self.dry_run:
                 self.log_info('Deletion of VPC skipped due to dry_run mode')
             else:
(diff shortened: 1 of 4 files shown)


ocw/lib/EC2.py
```diff
@@ -64,6 +64,14 @@ def ec2_client(self, region: str) -> "boto3.session.Session.client":
...
         return self.__ec2_client[region]
 
+    def eks_client(self, region: str) -> "boto3.session.Session.client":
+        if region not in self.__eks_client:
+            self.__eks_client[region] = boto3.client('eks',
+                                                     aws_access_key_id=self.auth_json['access_key'],
+                                                     aws_secret_access_key=self.auth_json['secret_key'],
+                                                     region_name=region)
+        return self.__eks_client[region]
+
```
pdostal, line 74: Is this needed?

ocw/lib/EC2.py
```diff
@@ -190,27 +198,16 @@ def cleanup_all(self) -> None:
         if PCWConfig.getBoolean('cleanup/vpc_cleanup', self._namespace):
             self.cleanup_vpcs()
 
-    def delete_elastic_ips(self, region):
```
pdostal, line 193: Is it causing issues?

ocw/lib/EC2.py
```diff
@@ -190,27 +198,16 @@ def cleanup_all(self) -> None:
...
-            self.delete_vpc_endpoints(region, vpc_id)
+            self.delete_routing_tables(region, vpc_id)
             self.delete_security_groups(vpc)
-            self.delete_vpc_peering_connections(region, vpc_id)
             self.delete_network_acls(vpc)
             self.delete_vpc_subnets(vpc)
+            self.delete_internet_gw(vpc)
+            self.delete_vpc_endpoints(region, vpc_id)
+            self.delete_vpc_peering_connections(region, vpc_id)
             if self.dry_run:
```
pdostal, line 211: Why is this here?

ricardobranco777 (approved):
LGTM

ocw/lib/EC2.py
asmorodskyi: because all functions which modify something on the cloud is guarded by dry_run variable so you have single switcher which makes PCW 100% harmless , which is very useful when we have some bug in code and PCW attempt to delete something what it should not . Or when new code is written and I have only "prod" environment to verify it

ocw/lib/EC2.py
asmorodskyi: with VPC deletion I never can say for 100% anything. The problem that VPC has plenty of different entities under control ( subnets , routing tables , routes , security groups , IPs , gateways and some others :) )  all they have some very complex relation between each other which I didn't manage fully realize. And if you trying to delete them in wrong order you will get exception. All changes which you see here is result of several days reading internet trigger different algorithms of deletion getting some exceptions and starting over again . 
So to answer you direct question **I don't know** but my gut feeling told me to drop it :D

tests/test_eks.py
```diff
@@ -94,18 +94,24 @@ def __init__(self, name, age):
...
+
     monkeypatch.setattr(PCWConfig, 'get_feature_property', mock_get_feature_property)
     monkeypatch.setattr(Provider, 'read_auth_json',
                         lambda *args, **kwargs: {'access_key': 'key', 'secret_key': 'secret'})
-    monkeypatch.setattr(EKS, 'list_regions', lambda self: ['region1'])
-    monkeypatch.setattr(Provider, 'cmd_exec', lambda *args, **kwargs: MockedSubprocessReturn(0))
+    monkeypatch.setattr(Provider, 'cmd_exec', mocked_cmd_exec)
     monkeypatch.setattr(EKS, 'aws_dir', lambda self: '/tmp')
 
     # monkeypatch.setattr(EC2, 'check_credentials', lambda *args, **kwargs: True)
```
ilausuch, line 117: I think this is a leftover, we could remove this

ilausuch (approved):
only a small change, but the code is ok. Great job!

ocw/lib/EC2.py
pdostal: I see, OK :+1:

ocw/lib/EC2.py
pdostal: But why is it called only after the other functions?

asmorodskyi: first we deleting all related resources and than we deleting vpc itself . 
if you asking why `if self.dry_run` not wrapping all function calls but only  exact call for deleting vpc - it is because each function internally has their own `if self.dry_run` checks . This way we doing everything what is possible w/o actually deleting resources , so when you have `dry_run = true ` you are closer to `dry_run = false` because you not skipping any querying calls but only deleting calls

pdostal: I got it now. Thank you

PR #240, opened 2023-04-11
Replace tabs with spaces

grisu48:
Align commands by replacing the tabs with spaces uniformly.


pdostal:
Go and add shellcheck.

pdostal (approved):

PR #244, opened 2023-04-20
Properly handle case where GatewayId is absent


asmorodskyi:
Properly handle case where GatewayId is absent


grisu48 (approved):

ricardobranco777 (approved):
LGTM

PR #246, opened 2023-04-28
Stop sending emails on VPC dependecy violation


asmorodskyi:
Stop sending emails on VPC dependecy violation


ocw/lib/EC2.py
```diff
@@ -164,7 +164,13 @@ def delete_vpc(self, region: str, vpc, vpc_id: str):
                 self.ec2_resource(region).meta.client.delete_vpc(VpcId=vpc_id)
             return None
         except Exception as ex:
-            return f"[{vpc_id}]{type(ex).__name__} on VPC deletion. {traceback.format_exc()}"
+            trace_txt = traceback.format_exc()
+            error_type = type(ex).__name__
+            del_responce = f"[{vpc_id}] {error_type} on VPC deletion. {trace_txt}"
+            self.log_err(del_responce)
+            if error_type == "ClientError" and "(DependencyViolation)" in trace_txt:
+                return None
```
grisu48, line 172: This is dangerous because the error doesn't get reported. Ignoring this error should not happen at this low level, but one level higher.

I would move this check down to lines 286, with a comment that we acknowledge this error but ignore it for now. But let's not ignore it at the lowest function level, where it's dangerous to swallow those error as "lie" to the function caller about success, when there is an actual error.

grisu48 (changes requested):
While I look forward to have those annoying notifications remove, I do believe we should move the ignore logic to the appropriate place.

ocw/lib/EC2.py
grisu48: If I'm offline due to connection issues today, consider the PR as approved, as soon as this issue is resolved. This is the only part I'm worried about.

pdostal:
Do we still need to run the Codecow locally as well? It's still mentioned in Makefile and maybe in some other places.

ocw/lib/EC2.py
asmorodskyi: let's say you have a script which suppose to delete empty folders on some FS , so you looping through all folders starting from root and try to delete folder using `rmdir` command . As you probably know `rmdir`  **throw exception**  when it try to delete folder which is none-empty , while technically it is exception should we really treat it that way in terms of task which I described ?  Here we have very similar situation `delete_vpc` is throwing exception when VPC is "not empty" .

Now about "not reporting error" . On current master we doing  two things when **any** exception happens : 
1. writing log message 
2. sending email 
Idea of the second one is to highlight some problem which might need immediate reaction from our side . e.g. when I see "CSP left overs" email with too long list of entities - it is signal to me to go and check up if cleanup is actually working . With this certain error **there is nothing what I can do** and it happening constantly due to bug in current VPC cleanup algorithm which certainly need improvements but it is different topic . But I can logging this exception into log so it is not fully lost

grisu48 (approved):
LGTM!

ocw/lib/EC2.py
asmorodskyi: I merged version which follows your advice , I personally like it less than what I did initially , so for now I merged it as separate commit so it is easy to revert . Based on what I said above I don't treat this exception as something what needs to be emailed . 
Another point - currently `delete_vpc` does not report if VPC deletion was successful or not . currently return value of `delete_vpc` answering on question "Was there some errors to report during deletion ?" and I really think we can safely answer "No there was not error **to report**" when `delete_vpc` throwing DependencyViolation error

grisu48:
Last nitpicking: `s/dependecy/dependency` in the title please.

asmorodskyi:
> dependency

most probably not the last I am not done here :) ( keep adding things )

PR #248, opened 2023-05-10
Add support for Openstack

ricardobranco777:
Add support for Openstack.

- Related ticket: https://progress.opensuse.org/issues/124577
- Related MR: https://gitlab.suse.de/qac/publiccloud-qa-suse-de/-/merge_requests/88


ocw/lib/openstack.py
```diff
@@ -0,0 +1,87 @@
...
+        servers = [vm for vm in self.client().compute.servers() if vm.startswith("openqa-vm-")]
+        self.log_dbg("Found {} servers", len(servers))
+        for server in servers:
+            if not self.is_outdated(parse(server.created_at)):
+                continue
+            if self.dry_run:
+                self.log_info("Instance termination {} skipped due to dry run mode", server.name)
+            else:
+                self.log_info("Deleting instance {}", server.name)
+                if not self.client().delete_server(
```
grisu48, line 52: According to the [documentation](https://docs.openstack.org/openstacksdk/latest/user/proxies/compute.html), this always returns `None`.

Following [some examples](https://docs.openstack.org/newton/user-guide/sdk-compute-apis.html#delete-server-api-v2) I couldn't find anything on how to detect errors here. Do you know a better way of doing this, while also being able to get errors?

grisu48 (commented):
One comment that we should address, otherwise LGTM!

ocw/lib/openstack.py
ricardobranco777: I'm using the `delete_server()` method of the `openstack.connection.Connection` class which returns a boolean value:

https://docs.openstack.org/openstacksdk/latest/user/connection.html#openstack.connection.Connection

I will add code to handle the OpenStackCloudException in every API call, though.

grisu48: Ahhh, I see. Thanks! Don't we need to catch `OpenStackCloudException` then?

grisu48 (approved):

ocw/lib/openstack.py
```diff
@@ -0,0 +1,108 @@
...
+        max_days = PCWConfig.get_feature_property('cleanup', 'openstack-max-age-days', self._namespace)
+        now = datetime.now(timezone.utc)
+        try:
+            images = [image for image in self.client().image.images() if "openqa" in image.tags]
+        except OpenStackCloudException as exc:
+            self.log_warn("Got exception while listing images: {}", exc)
+            return
+        self.log_dbg("Found {} images", len(images))
+        for image in images:
+            if (now - parse(image.created_at)).days > max_days:
```
pdostal, line 78: Why don't you use `is_outdated` as in servers?

pdostal (approved):
Fine from me. Just, will we also wait for Anton?

ocw/lib/openstack.py
```diff
@@ -0,0 +1,108 @@
...
+                region_name=self.get_data('region_name'),
+                user_domain_name=self.get_data('user_domain_name'),
+                project_domain_name=self.get_data('project_domain_id'),
+            )
+        return self.__client
+
+    def cleanup_all(self) -> None:
+        self._cleanup_instances()
+        self._cleanup_images()
+        self._cleanup_keypairs()
```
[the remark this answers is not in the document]
asmorodskyi: while I do see 100% code coverage as our goal there are some lines which more important to cover and there are others which just "nice to have " . this function is second one , because only thing what this function is doing is calling other functions so only test which we can do "does it call this 3 functions ?" which IMHO just "nice to have"

ocw/lib/openstack.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def _cleanup_instances(self) -> None:
+        # Delete VM's & associated floating IP address(es)
+        try:
+            servers = [vm for vm in self.client().compute.servers() if vm.name.startswith("openqa-vm-")]
```
asmorodskyi, line 45: do we really need this kind of limitation ? why we want to limit our self's here ?

asmorodskyi: why deletion of VM's is not obeying `dry_run` option ?

asmorodskyi: we not checking age of VM is it intentional ?

asmorodskyi: sorry I still not fully functioning >_< ofc. you just collecting servers and not actually deleting them so please ignore second and 3d questions

requirements.txt
```diff
@@ -15,6 +15,7 @@ texttable
 oauth2client
 google-api-python-client==2.55.0
 google-cloud-storage==2.4.0
+openstacksdk~=1.1.0
```
asmorodskyi, line 18: just to double-check `~=` means "upgrade minor versions " , right ?

ocw/lib/cleanup.py
```diff
@@ -25,6 +26,9 @@ def cleanup_run():
             if ProviderChoice.GCE in providers:
                 GCE(namespace).cleanup_all()
 
+            if ProviderChoice.Openstack in providers:
```
asmorodskyi, line 29: looks like you forgot to update enum with new entry ( at least I fail to find this in PR )

asmorodskyi:
Well done Ricardo ! You manage to fully follow codestyle of this project . The ONLY thing which I would like to clarify is about calculation of is_outdated for entities

ricardobranco777:
> This is very nice! Have you actually tried to run it without dry-run mode? I am curious to see if you can delete an image or keypair which is used by a running VM.  

Not yet. I will when I have time.

ocw/lib/openstack.py
ricardobranco777: For VM's isn't 1 day enough?
For images it's configurable and 3 to 4 days is better to restart tests.

ricardobranco777: > Why don't you use `is_outdated` as in servers?

You're right. It must be the same calculation done with VM's.

asmorodskyi: we not discussing exact value here , we should have ability to tweak over a time when some requirements change . Only thing we discuss right now are we gonna do it via `max-age-days` or `openstack-max-age-days`  or  we keep some custom calculation within the code ( less preferable because than you can not actually tweak it )  ?

ricardobranco777: > we not discussing exact value here , we should have ability to tweak it over a time when some requirements change . Only thing we discuss right now are we gonna do it via `max-age-days` or `openstack-max-age-days` or we keep some custom calculation within the code ( less preferable because than you can not actually tweak it ) ?

I prefer than for Openstack we have a different setting.

Also, do we actually see VM's running for more than one day?  For images the setting makes sense because tests can be restarted.

ocw/lib/openstack.py
```diff
@@ -0,0 +1,108 @@
...
+        for keypair in keypairs:
+            if self.dry_run:
+                self.log_info("Keypair deletion {} skipped due to dry run mode", keypair.name)
+            else:
+                self.log_info("Deleting keypair {}", keypair.name)
+                try:
+                    if not self.client().delete_keypair(keypair.name):
+                        self.log_err("Failed to delete keypair {}", keypair.name)
+                except OpenStackCloudException as exc:
+                    self.log_warn("Got exception while deleting keypair {}: {}", keypair.name, exc)
```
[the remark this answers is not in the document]
ricardobranco777: > Why don't we check the key age over here?

The `openstack.compute.v2.keypair.Keypair` does have a `created_at` attribute but our server is not populating it and the Terraform provider doesn't have the capability to add tags to keypairs.

So I'm inclined to remove the whole `_cleanup_keypairs` method and instead of creating a new keypair for each test as done [here](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/publiccloud/terraform/openstack.tf#L26-L30) use a manually created keypair as documented in:
https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/compute_keypair_v2

@asmorodskyi @jlausuch WDYT?

ocw/lib/openstack.py
ricardobranco777: Our VM's start with "openqa-vm-".  Do you want to make it configurable?  We _have_ to limit ourselves here because we only want to destroy VM's created by openQA.

requirements.txt
ricardobranco777: > just to double-check `~=` means "upgrade minor versions " , right ?

Yes.

ocw/lib/openstack.py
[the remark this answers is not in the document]
asmorodskyi: I see to separate discussions which I don't want to mix into one . Only important point from PCW POV is that we don't want to clean up key-pairs because there is no way to easily understand age of certain key . Did I get this part right ? 

The part about how to handle key creation within openQA tests let's do in corresponding PR to openQA repo

ocw/lib/openstack.py
asmorodskyi: why ? isn't this cluster used ONLY for openQA testing ? who else using it and for what ?  From experience which I have for cleanup VMs in Azure/AWS/GCE . At the beginning we also used this super safe approach where we delete only VMs with certain pattern in the name but than it appears that we have sooo many cases where because of this I need to manually cleanup VMs which name is not following this pattern because of the "reasons".  So after all I decided to drop this extra filter because anyway 99% VMs which are created in our accounts suppose to be cleaned up and for 1% we have "pcw_ignore"  tag which can protect them .

ocw/lib/openstack.py
asmorodskyi: > Also, do we actually see VM's running for more than one day?

I just want to have flexibility to give different answers to this question and not to have hardcoded value

ocw/lib/openstack.py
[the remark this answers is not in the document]
ricardobranco777: > I see to separate discussions which I don't want to mix into one . Only important point from PCW POV is that we don't want to clean up key-pairs because there is no way to easily understand age of certain key . Did I get this part right ?
> 

Yes.

> The part about how to handle key creation within openQA tests let's do in corresponding PR to openQA repo

Agree.  I will remove the whole `cleanup_keypairs()` stuff.

ocw/lib/openstack.py
ricardobranco777: These non-openQA VM's are running right now:

`['fgerling-ankh', 'fgerling-retrospected', 'SLE 15-SP3 GMC', 'yifan0']`

I don't think it's wise to push the `pcw_ignore` tag for Openstack users.  We only pay for the electricity here, so this safe approach makes sense, IMO.

ocw/lib/openstack.py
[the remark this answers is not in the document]
ricardobranco777: I have to call `client.compute.get_keypair(keypair.name)` for each keypair to get the additional info

ricardobranco777:
> This is very nice! Have you actually tried to run it without dry-run mode? I am curious to see if you can delete an image or keypair which is used by a running VM.

Quoting this [page](https://platform9.com/kb/openstack/deleting-images-used-to-create-running-instances):

_OpenStack will not prevent deletion of an image which was used to create an instance that is still in an active state. However, if the image is deleted, it will not be possible to migrate the instance successfully. To prepare for deleting an image for which instances still exist that were created from it, it will be necessary to create a snapshot of each instance and recreate the instance using the snapshot. This will allow migration to function as expected._

PR #250, opened 2023-05-20
Align variable naming in openQA and PCW


asmorodskyi:
Align variable naming in openQA and PCW


pdostal (approved):
Does this mean that we had to have both keys in our secret JSON?

asmorodskyi:
> Does this mean that we had to have both keys in our secret JSON?

currently key used by PCW located under `/srv/pcw/creds/<namespace>` and key used by openQA test is under `/srv/www/secrets/<namespace>`   so secrets physically in different files anyway . And currently for AWS only they have different field names . I wanted to align this since this difference was introduced but didn't have time for this till now

PR #255, opened 2023-05-28
Fix authentication issue with Openstack

ricardobranco777:
This PR:
- Adds debug support for Openstack
- Fixes authentication issue with Openstack

### DIFF
--- a/ocw/lib/openstack.py
+++ b/ocw/lib/openstack.py
@@ -4,6 +4,7 @@ from dateutil.parser import parse
 import openstack
 from openstack.exceptions import OpenStackCloudException
 from webui.PCWConfig import PCWConfig
+from webui.settings import DEBUG
 from .provider import Provider
 
 
@@ -23,6 +24,7 @@ class Openstack(Provider):
     def client(self) -> None:
         if self.__client is None:
             self.__client = openstack.connect(
+                debug=DEBUG,
                 insecure=True,  # Trust the certificate
                 auth_url=self.get_data('auth_url'),
                 project_name=self.get_data('project_name'),
--- a/ocw/lib/openstack.py
+++ b/ocw/lib/openstack.py
@@ -32,7 +32,9 @@ class Openstack(Provider):
                 password=self.get_data('password'),
                 region_name=self.get_data('region_name'),
                 user_domain_name=self.get_data('user_domain_name'),
-                project_domain_name=self.get_data('project_domain_id'),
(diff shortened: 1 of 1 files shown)


jlausuch (approved):

ocw/lib/openstack.py
```diff
@@ -23,14 +24,17 @@ def __new__(cls, vault_namespace: str):
     def client(self) -> None:
         if self.__client is None:
             self.__client = openstack.connect(
+                debug=DEBUG,
```
asmorodskyi, line 27: I am not sure that it is good idea to keep it constantly on DEBUG , it can spam logs a lot . Instead we defining some common log level for whole app here https://github.com/SUSE/pcw/blob/master/templates/pcw.ini#L11 what if you will use this one ? Or at least define dedicated variable in `pcw.ini` for it . I want to have ability to mute it w/o tweaking the code

ocw/lib/openstack.py
```diff
@@ -23,14 +24,17 @@ def __new__(cls, vault_namespace: str):
...
                 insecure=True,  # Trust the certificate
                 auth_url=self.get_data('auth_url'),
                 project_name=self.get_data('project_name'),
                 username=self.get_data('username'),
                 password=self.get_data('password'),
                 region_name=self.get_data('region_name'),
                 user_domain_name=self.get_data('user_domain_name'),
-                project_domain_name=self.get_data('project_domain_id'),
+                project_id=self.get_data('project_id'),
+                load_envvars=False,  # Avoid reading OS_* environment variables
```
asmorodskyi, line 36: can elaborate why both needs to be avoided ? ( I am not against just feel that there is some story behind which I want to know ... )

ricardobranco777: The original code worked on my machine because I had those environment variables turned on, but I was assigning `project_domain_id` to `project_domain_name` and our Openstack instance was getting info from a SUMA project for some reason (maybe because it's default).

ocw/lib/openstack.py
ricardobranco777: It can be muted w/o tweaking the code because `debug` is only `True` if set here:
https://github.com/SUSE/pcw/blob/master/webui/settings.py#L30

ocw/lib/openstack.py
asmorodskyi: ok

PR #264, opened 2023-06-14
gce: Centralize deletion & pagination logic

ricardobranco777:
This PR:
- Centralizes deletion & pagination logic.
- Upgrades to Django 4.2.x (which fixes pytest-django's DeprecationWarning).
- Adds typing hints.
- Use f-string's where applicable.
- Minor refactor.

I need this PR merged before adding code to remove firewall-rules, routes, subnets & networks without repeating the deletion & pagination logic.

### DIFF
--- a/ocw/lib/gce.py
+++ b/ocw/lib/gce.py
@@ -91,15 +91,9 @@ class GCE(Provider):
         return [GCE.url_to_name(z) for z in region["zones"]]
 
     def delete_instance(self, instance_id, zone) -> None:
-        if self.dry_run:
-            self.log_info(
-                "Deletion of instance {} skipped due to dry run mode", instance_id
-            )
-        else:
-            self.log_info("Delete instance {}".format(instance_id))
-            self.compute_client().instances().delete(
-                project=self.project, zone=zone, instance=instance_id
-            ).execute()
+        self._delete_resource(
+            self.compute_client().instances, instance_id, project=self.project, zone=zone, instance=instance_id
+        )
 
     @staticmethod
     def url_to_name(url) -> str:
--- a/ocw/lib/gce.py
+++ b/ocw/lib/gce.py
@@ -1,4 +1,5 @@
 import json
+from os.path import basename
 from datetime import timezone
 from dateutil.parser import parse
 import googleapiclient.discovery
@@ -88,16 +89,12 @@ class GCE(Provider):
             .get(project=self.project, region=region)
             .execute()
         )
-        return [GCE.url_to_name(z) for z in region["zones"]]
+        return [basename(z) for z in region["zones"]]
 
     def delete_instance(self, instance_id, zone) -> None:
         self._delete_resource(
             self.compute_client().instances, instance_id, project=self.project, zone=zone, instance=instance_id
         )
-
-    @staticmethod
-    def url_to_name(url) -> str:
-        return url[url.rindex("/")+1:]
 
     @staticmethod
     def get_error_reason(error: "googleapiclient.errors.HttpError") -> str:
--- a/ocw/lib/gce.py
+++ b/ocw/lib/gce.py
@@ -126,44 +126,15 @@ class GCE(Provider):
                 self.log_dbg(f"{len(disks)} disks found")
                 for disk in disks:
                     if self.is_outdated(parse(disk["creationTimestamp"]).astimezone(timezone.utc)):
-                        if self.dry_run:
-                            self.log_info("Deletion of disk {} created on {} skipped due to dry run mode",
-                                          disk["name"], disk["creationTimestamp"])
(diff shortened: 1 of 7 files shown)


mpagot (approved):
LGTM

Makefile
```diff
@@ -14,11 +14,11 @@ test:
 	flake8 --max-line-length=130 ocw
 	flake8 --max-line-length=130 manage.py
 	flake8 --max-line-length=130 cleanup_k8s.py
-	pytest --cov
+	pytest --cov -W ignore::DeprecationWarning
```
asmorodskyi, line 17: I am not sure that this is good idea :thinking:  maybe we need to follow warnings and stop using what is deprecated instead ? use of deprecated API is increase of technical debt which you will need to pay anyway one day ...

ricardobranco777: > I am not sure that this is good idea thinking maybe we need to follow warnings and stop using what is deprecated instead ? use of deprecated API is increase of technical debt which you will need to pay anyway one day ...

The deprecated warning comes from Django 4.0.x.

Either we update Django or use this temporary workaround, because pytest refuses to run without it.

ocw/lib/gce.py
```diff
@@ -21,58 +21,45 @@ def __init__(self, namespace):
...
+        return self._paginated(self.compute_client().instances, project=self.project, zone=zone)
 
-    def list_all_instances(self):
+    def list_all_instances(self) -> list:
         result = []
         self.log_dbg("Call list_all_instances")
         for region in self.list_regions():
             for zone in self.list_zones(region):
-                result += self.list_instances(zone=zone)
+                result.extend(self.list_instances(zone=zone))
```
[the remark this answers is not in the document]
ricardobranco777: > I am not against that but would like to hear why extend is better than `+=` ?

I read somewhere that `+=` creates and assigns a new list while `.extend()` uses the memory of the extended list instead, thus being faster.

ocw/lib/gce.py
```diff
@@ -106,77 +93,54 @@ def get_error_reason(error: "googleapiclient.errors.HttpError") -> str:
...
-                                    if GCE.get_error_reason(err) == 'resourceInUseByAnotherResource':
-                                        self.log_dbg("Disk {} can not be deleted because in use", disk["name"])
-                                    else:
-                                        raise err
-                    request = (
-                        self.compute_client()
-                        .disks()
-                        .list_next(previous_request=request, previous_response=response)
-                    )
+                disks = self._paginated(self.compute_client().disks, project=self.project, zone=zone)
```
asmorodskyi, line 103: you lost debug message here `self.log_dbg("response has no items. id={}", response["id"])` . you still can add something like : 
`self.log_dbg("{} disks found", len(disks))`

ocw/lib/gce.py
```diff
@@ -21,58 +21,45 @@ def __init__(self, namespace):
...
 
-    def compute_client(self):
+    @staticmethod
+    def _paginated(api_call, **kwargs) -> list:
+        results = []
+        request = api_call().list(**kwargs)
+        while request is not None:
+            response = request.execute()
+            if "items" in response:
+                results.extend(response["items"])
```
asmorodskyi, line 31:
```
else: 
   self.log_dbg("response has no items. id={}", response["id"])
```

tests/test_gce.py
```diff
@@ -73,7 +73,9 @@ def delete(self, *args, **kwargs):
             self.deleted_disks.append(kwargs['disk'])
         else:
             raise ValueError("Unexpected delete request")
-        return self.responses.pop(0)
+        if len(self.responses) > 0:
```
asmorodskyi, line 76: nice !

Makefile
asmorodskyi: than we need to update Django !

ocw/lib/gce.py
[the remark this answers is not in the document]
asmorodskyi: thanks for clarification

asmorodskyi:
really nice PR , I find new code looks more readable and compact !

asmorodskyi (changes requested):
blocking due to DeprecationWarning ignorance . we need to update Django and remove ignore of DeprecationWarning

webui/settings.py
```diff
@@ -128,6 +128,7 @@
 
 USE_I18N = True
 
+# DO NOT CHANGE THIS VALUE!
```
asmorodskyi, line 131: maybe few words about WHY it is bad idea to change it ?

ricardobranco777: Done.

pdostal (approved):

asmorodskyi (approved):
LGTM

PR #266, opened 2023-06-16
Add missed data migration


asmorodskyi:
Add missed data migration


asmorodskyi:
Looks like it is something what was missed in #255

ricardobranco777 (approved):
LGTM

PR #269, opened 2023-06-19
gce: Refactor test code

ricardobranco777:
Refactor test code for GCE and add more tests.

This will make it easier to add new tests for upcoming functionality.

### DIFF
--- a/tests/test_gce.py
+++ b/tests/test_gce.py
@@ -1,11 +1,13 @@
+from pytest import fixture
+from unittest.mock import MagicMock, patch
 import json
 from datetime import datetime, timezone, timedelta
-from ocw.lib.gce import GCE, Provider
+from ocw.lib.gce import GCE
 from webui.PCWConfig import PCWConfig
 from tests.generators import max_age_hours, mock_get_feature_property
 
 
-class FakeRequest:
+class MockRequest:
     def __init__(self, response={}):
         self.response = response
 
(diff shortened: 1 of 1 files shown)


tests/test_gce.py
```diff
@@ -1,27 +1,44 @@
...
+        self.deleted_resources = list()
         self.responses = responses
 
+    def __call__(self, *args, **kwargs):
+        return self
+
     def list(self, *args, **kwargs):
-        return self.responses.pop(0)
+        if len(self.responses) > 0:
+            return self.responses.pop(0)
```
ilausuch, line 28: we don't need this, but I understand if we have >0 could be more than one, so pop all of them?

ilausuch (commented):
Looks good!

tests/test_gce.py
```diff
@@ -1,143 +1,165 @@
...
-            raise ValueError("Unexpected delete request")
-        if len(self.responses) > 0:
-            return self.responses.pop(0)
-        return FakeRequest(None)
+
+def test_list_all_instances(gce):
+    instances = MockResource([MockRequest({'items': ['instance1', 'instance2']}), None])
+    gce.compute_client.instances = instances
+    with (
+        patch.object(gce, "list_regions", lambda *args, **kwargs: ["region1"]),
```
asmorodskyi, line 83: maybe this should be fixture ? it is repeated several times

tests/test_gce.py
```diff
@@ -1,143 +1,165 @@
...
+        gce.compute_client.instances = lambda *args, **kwargs: []
+        setattr(gce.compute_client, resource_type, resources)
+
+        if dry_run:
+            gce.dry_run = True
+            cleanup_call()
+            assert resources.deleted_resources == []
+        else:
+            cleanup_call()
+            assert resources.deleted_resources == ['delete1', 'delete2']
```
asmorodskyi, line 134:
```suggestion
      gce.dry_run = True
      cleanup_call()
      assert resources.deleted_resources == []
       cleanup_call()
        assert resources.deleted_resources == ['delete1', 'delete2']
```

anyway for every entity you covering both cases , so just unconditionally cover both case and merge all `test_cleanup_<entity>` and `test_cleanup_<entity>_dry_run` into single test

ricardobranco777: > anyway for every entity you covering both cases , so just unconditionally cover both case and merge all `test_cleanup_<entity>` and `test_cleanup_<entity>_dry_run` into single test

I had to use a loop because I can't reuse `resources` .

asmorodskyi: ok

PR #270, opened 2023-06-21
gce: Also delete GCE network resources

ricardobranco777:
This PR deletes these letf-over network resources on GCE: firewall rules, routes, subnetworks & networks, with care taken not to delete any resource associated with the `default` network.

Current networks in ei-sle-qa: default, networktest1 & qe-c-testing-k8s-network.

Related ticket: https://progress.opensuse.org/issues/127220

### DIFF
--- a/ocw/lib/gce.py
+++ b/ocw/lib/gce.py
@@ -10,6 +10,7 @@ from .provider import Provider
 
 class GCE(Provider):
     __instances = {}
+    __skip_networks = frozenset({"default"})
 
     def __new__(cls, vault_namespace):
         if vault_namespace not in GCE.__instances:
@@ -36,9 +37,13 @@ class GCE(Provider):
 
     def _delete_resource(self, api_call, resource_name, *args, **kwargs) -> None:
         resource_type = {
-            self.compute_client().instances: "instance",
-            self.compute_client().images: "image",
             self.compute_client().disks: "disk",
+            self.compute_client().firewalls: "firewall",
+            self.compute_client().images: "image",
+            self.compute_client().instances: "instance",
+            self.compute_client().networks: "network",
+            self.compute_client().routes: "route",
+            self.compute_client().subnetworks: "subnetwork",
         }.get(api_call, "resource")
         if self.dry_run:
             self.log_info(f"Deletion of {resource_type} {resource_name} skipped due to dry run mode")
@@ -110,6 +115,10 @@ class GCE(Provider):
         self.log_info("Call cleanup_all")
         self.cleanup_disks()
         self.cleanup_images()
+        self.cleanup_firewalls()
+        self.cleanup_routes()
+        self.cleanup_subnetworks()
+        self.cleanup_networks()
 
     def cleanup_disks(self) -> None:
         self.log_dbg("Disks cleanup")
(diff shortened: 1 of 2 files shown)


ocw/lib/gce.py
```diff
@@ -10,6 +10,8 @@
 
 class GCE(Provider):
     __instances = {}
+    # XXX: Perhaps we should get this from pcw.ini
+    __skip_networks = frozenset({"default"})
```
mpagot, line 14: how this one works?

ocw/lib/gce.py
```diff
@@ -36,9 +38,13 @@ def _paginated(self, api_call, **kwargs) -> list:
 
     def _delete_resource(self, api_call, resource_name, *args, **kwargs) -> None:
         resource_type = {
-            self.compute_client().instances: "instance",
-            self.compute_client().images: "image",
             self.compute_client().disks: "disk",
+            self.compute_client().firewalls: "firewall",
```
mpagot, line 42: does the order matter here?

ocw/lib/gce.py
ricardobranco777: > how this one works?

Testing for membership is faster on sets, and immutables sets are faster than vanilla sets.

mpagot: so this one is a sort of dummy initialization for performance in some following part of code. Right? Has TTL value from pcw.ini to be taken in account here or later? Which is following part of code involved? Is `is_outdated` ?

ocw/lib/gce.py
mpagot: Does not some resources to be deleted before than others as they are inter-dependent ?

ocw/lib/gce.py
```diff
@@ -10,6 +10,8 @@
 
 class GCE(Provider):
     __instances = {}
+    # XXX: Perhaps we should get this from pcw.ini
```
asmorodskyi, line 13: 1. please delete the comment  , no point to keep it in the code 
2. I would say that from beginning it is fine to keep it hardcoded . If there will be need to tweak it we will analyze this need and decide about moving it to pcw.ini

ricardobranco777: The XXX was added for discussion.

Fixed.

ocw/lib/gce.py
ricardobranco777: This value is only for skipping networks in cleanup_xxx(), no TTL involved.

ocw/lib/gce.py
ricardobranco777: This is a dictionary to get the resource type name from the API method.

We clean up the resources in `cleanup_all()` in order.  Check the force-pushed commit.

ocw/lib/gce.py
asmorodskyi: DISCLAIMER : I am not against use of immutable sets here . I just want to raise more global topic not directly related to this PR .

> Testing for membership is faster on sets, and immutables sets are faster than vanilla sets.

This argument can not be used as a prove of something in context of PCW . We speaking about single threaded app which is doing one iteration in 20 minutes and set which we discussing will contain no more than 100 elements in very very bad situation . 
Such argument can be used in case of multi-threaded up which processing 10000000 elements in several threads at the same time. 
Let's always keep context of what we doing and why . Again maybe not directly related to this certain case but I have seen so many cases when people were spending few days to improve performance of test and make it run 1 hour 2 minutes instead of 1 hour 10 minutes .... 

TDLR Performance of code execution is certainly not the top prio for PCW and for most of projects which we maintaining ...

asmorodskyi (approved):
nice !

mpagot (approved):
LGTM

ocw/lib/gce.py
```diff
@@ -76,12 +82,14 @@ def list_all_instances(self) -> list:
                 result.extend(self.list_instances(zone=zone))
         return result
 
+    # XXX: Perhaps we should cache this with a TTL?
```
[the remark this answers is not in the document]
asmorodskyi: this call takes approx. 1 minute I assume . IMO 1 minute is fair price to keep code simpler . Introduction of cache will with several extra scenarios how PCW  can fail . 
Another  point that all cleanup logic is happening in container which is destroyed after each execution so cache should be some real persistence logic which will dump it into file or some DB

ocw/lib/gce.py
ricardobranco777: > TDLR Performance of code execution is certainly not the top prio for PCW and for most of projects which we maintaining ...

I was about to add a `concurrent.futures.ThreadPoolExecutor()` to the delete calls for each region.

I'm joking.  XD

PR #271, opened 2023-06-21
eks: Delete the services associated with the cluster

ricardobranco777:
This PR:
- Deletes the services associated with a cluster in EKS
- Minor code tidying to avoid indentation & f-strings.

Related ticket: https://progress.opensuse.org/issues/127268

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,15 @@
 CONT_TAG=suse/qac/pcw
 
 .PHONY: all
-all: prepare test
+all: prepare test pylint
 
 .PHONY: prepare
 prepare:
 	pip install -r requirements_test.txt
+
+.PHONY: pylint
+pylint:
+	pylint ocw/lib/*.py cleanup_k8s.py
 
 .PHONY: test
 test:
--- a/ocw/lib/eks.py
+++ b/ocw/lib/eks.py
@@ -86,15 +86,15 @@ class EKS(Provider):
     def all_clusters(self) -> dict:
         clusters = {}
         for region in EKS.__cluster_regions:
-            self.log_dbg("Checking clusters in {}", region)
+            self.log_dbg(f"Checking clusters in {region}")
             response = self.eks_client(region).list_clusters()
             if 'clusters' in response and len(response['clusters']) > 0:
                 clusters[region] = []
-                self.log_dbg("Found {} clusters in {}", len(response['clusters']), region)
+                self.log_dbg(f"Found {len(response['clusters'])} clusters in {region}")
                 for cluster in response['clusters']:
                     cluster_description = self.eks_client(region).describe_cluster(name=cluster)
                     if 'cluster' not in cluster_description or 'tags' not in cluster_description['cluster']:
-                        self.log_err("Unexpected cluster description: {}", cluster_description)
+                        self.log_err(f"Unexpected cluster description: {cluster_description}")
                     elif TAG_IGNORE not in cluster_description['cluster']['tags']:
                         clusters[region].append(cluster)
                 if len(clusters[region]) == 0:
@@ -105,7 +105,7 @@ class EKS(Provider):
         if self.dry_run:
             self.log_info("Skip waiting due to dry-run mode")
             return None
-        self.log_dbg("Waiting empty nodegroup list in {}", cluster_name)
+        self.log_dbg(f"Waiting empty nodegroup list in {cluster_name}")
         end = datetime.now(timezone.utc) + timedelta(minutes=timeout_minutes)
         resp_nodegroup = self.eks_client(region).list_nodegroups(clusterName=cluster_name)
 
@@ -113,7 +113,7 @@ class EKS(Provider):
             time.sleep(20)
             resp_nodegroup = self.eks_client(region).list_nodegroups(clusterName=cluster_name)
             if len(resp_nodegroup['nodegroups']) > 0:
-                self.log_dbg("Still waiting for {} nodegroups to disappear", len(resp_nodegroup['nodegroups']))
+                self.log_dbg(f"Still waiting for {len(resp_nodegroup['nodegroups'])} nodegroups to disappear")
         return None
 
     def delete_all_clusters(self) -> None:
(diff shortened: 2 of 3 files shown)


ilausuch (approved):

ocw/lib/eks.py
```diff
@@ -86,15 +86,15 @@ def kubectl_client(self, region: str, cluster_name: str):
     def all_clusters(self) -> dict:
         clusters = {}
         for region in EKS.__cluster_regions:
-            self.log_dbg("Checking clusters in {}", region)
+            self.log_dbg(f"Checking clusters in {region}")
```
asmorodskyi, line 89: I would like to use same style everywhere , if you want to switch to f-strings please prepare commit which do it everywhere not just in few files

ricardobranco777: > I would like to use same style everywhere , if you want to switch to f-strings please prepare commit which do it everywhere not just in few files

Will do it in this PR then.

ricardobranco777: > I would like to use same style everywhere , if you want to switch to f-strings please prepare commit which do it everywhere not just in few files

There are far too many lines to change for me to make that change right now.  In some instances the line will exceed the line limit of 130 chars.  So either we do it one module at a time or I rollback the f-string changes in this file.

ricardobranco777: > I would like to use same style everywhere , if you want to switch to f-strings please prepare commit which do it everywhere not just in few files

These Kubernetes scripts already use f-strings:
- [https://github.com/SUSE/pcw/blob/master/ocw/lib/aks.py](ocw/lib/aks.py)
- [https://github.com/SUSE/pcw/blob/master/ocw/lib/gke.py](ocw/lib/gke.py)
- [https://github.com/SUSE/pcw/blob/master/ocw/lib/k8s.py](ocw/lib/k8s.py)

GCE & Openstack also use them.  I promise to change EC2 & Azure when I touch them.

asmorodskyi: ok

ocw/lib/eks.py
```diff
@@ -105,40 +105,49 @@ def wait_for_empty_nodegroup_list(self, region: str, cluster_name: str, timeout_
...
-                                self.log_info("Skipping {} nodegroup deletion due to dry-run mode", nodegroup)
-                            else:
-                                self.log_info("Deleting {}", nodegroup)
-                                self.eks_client(region).delete_nodegroup(
-                                    clusterName=cluster, nodegroupName=nodegroup)
-                        self.wait_for_empty_nodegroup_list(region, cluster)
+                self.log_dbg(f"Found {len(response['clusters'])} cluster(s) in {region}")
+            for cluster in response['clusters']:
+                resp_nodegroup = self.eks_client(region).list_nodegroups(clusterName=cluster)
+                if len(resp_nodegroup['nodegroups']):
```
grisu48, line 127:
```suggestion
                if 'nodegroups' in resp_nodegroup:
```
I think it's better to check for the key in the dict, otherwise we get a `KeyError` if the item is not present.

ocw/lib/eks.py
```diff
@@ -105,40 +105,49 @@ def wait_for_empty_nodegroup_list(self, region: str, cluster_name: str, timeout_
...
+                    self.log_dbg(f"Found {len(resp_nodegroup['nodegroups'])} nodegroups for {cluster}")
+                for nodegroup in resp_nodegroup['nodegroups']:
+                    if self.dry_run:
+                        self.log_info(f"Skipping {nodegroup} nodegroup deletion due to dry-run mode")
+                    else:
+                        self.log_info(f"Deleting {nodegroup}")
+                        self.eks_client(region).delete_nodegroup(
+                            clusterName=cluster, nodegroupName=nodegroup)
+                if len(resp_nodegroup['nodegroups']):
```
grisu48, line 136:
```suggestion
                if 'nodegroups' in resp_nodegroup:
```

pylintrc
```diff
@@ -1,5 +1,6 @@
 [MASTER]
 disable=
+ C0103, # Module name "EC2" doesn't conform to snake_case naming style
```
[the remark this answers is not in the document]
asmorodskyi: this applies to all additions below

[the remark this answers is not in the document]
ricardobranco777: I too disagree, but then we should change the approach to pylint and don't ignore most stuff that is already there.

Makefile
```diff
@@ -8,6 +8,10 @@ all: prepare test
 prepare:
 	pip install -r requirements_test.txt
 
+.PHONY: pylint
```
asmorodskyi, line 11: maybe add this to all ?

pylintrc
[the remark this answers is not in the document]
ricardobranco777: Fixed.

Makefile
ricardobranco777: Fixed.

ricardobranco777:
> I muted some errors in linter because at the point when I was adding linter to project to fix ALL complains we would need to change too big places in code . So I just "make it green" but from current state there is NO options to add more exceptions . For sure not in batch mode . Maybe we can add one or two to this ignore list with really really good explanation why we don't want to obey this rule. I am actually expecting work on REMOVAL lines from this file not adding more things to ignore list

It went from 9.95 to 9.94 because of `ocw/lib/eks.py:119:4: R0912: Too many branches (13/12) (too-many-branches)`

The only way to avoid it would be:

```
-                if 'nodegroups' in resp_nodegroup:
-                    self.log_dbg(f"Found {len(resp_nodegroup['nodegroups'])} nodegroups for {cluster}")
+                self.log_dbg(f"Found {len(resp_nodegroup['nodegroups'])} nodegroups for {cluster}")
```

asmorodskyi:
> The only way to avoid it would be:
> 
> ```
> -                if 'nodegroups' in resp_nodegroup:
> -                    self.log_dbg(f"Found {len(resp_nodegroup['nodegroups'])} nodegroups for {cluster}")
> +                self.log_dbg(f"Found {len(resp_nodegroup['nodegroups'])} nodegroups for {cluster}")
> ```


https://pycodequ.al/docs/pylint-messages/r0912-too-many-branches.html , in how to fix it there are lines : 
```
Usually, functions with too many branches fall in one of two categories: (A) functions that do more than one thing or (B) functions that are just a map lookup.
```

I really think that our case is (A) . And actually even without a linter I saw the need to split this big function into smaller ones . e.g. deleting nodes should be separate function . 

P.S. This exactly why I find linter useful and like to follow it complains . Each check didn't come from the void , each check is result of some previous cases where code with certain patterns produced certain problems . So when you get complain from linter it is usually good advice how to improve your code

ricardobranco777:
> > The only way to avoid it would be:
> > ```
> > -                if 'nodegroups' in resp_nodegroup:
> > -                    self.log_dbg(f"Found {len(resp_nodegroup['nodegroups'])} nodegroups for {cluster}")
> > +                self.log_dbg(f"Found {len(resp_nodegroup['nodegroups'])} nodegroups for {cluster}")
> > ```
> 
> https://pycodequ.al/docs/pylint-messages/r0912-too-many-branches.html , in how to fix it there are lines :
> 
> ```
> Usually, functions with too many branches fall in one of two categories: (A) functions that do more than one thing or (B) functions that are just a map lookup.
> ```
> 
> I really think that our case is (A) . And actually even without a linter I saw the need to split this big function into smaller ones . e.g. deleting nodes should be separate function .
> 
> P.S. This exactly why I find linter useful and like to follow it complains . Each check didn't come from the void , each check is result of some previous cases where code with certain patterns produced certain problems . So when you get complain from linter it is usually good advice how to improve your code

Fixed.  I refactored `delete_all_clusters()` and created `delete_nodegrups()` & `delete_services()` along with their respective test functions.

ocw/lib/eks.py
```diff
@@ -105,40 +105,54 @@ def wait_for_empty_nodegroup_list(self, region: str, cluster_name: str, timeout_
         if self.dry_run:
             self.log_info("Skip waiting due to dry-run mode")
             return None
-        self.log_dbg("Waiting empty nodegroup list in {}", cluster_name)
+        self.log_dbg(f"Waiting empty nodegroup list in {cluster_name}")
```
asmorodskyi, line 108: interesting , this means that when test is running log level is not DEBUG ? :thinking:  Maybe we can tweak log level and increase coverage ? :D

asmorodskyi (approved):
Perfecto !

PR #274, opened 2023-07-02
Minor fixes

ricardobranco777:
This PR:
- Fixes an issue with our Azure code getting ResourceManagementClient each time `resource_mgmt_client()` is called.
- Uses f-strings everywhere now.


ocw/lib/azure.py
```diff
@@ -137,7 +135,7 @@ def container_valid_for_cleanup(container) -> bool:
         '''
         if Instance.TAG_IGNORE in container['metadata']:
             return False
-        if re.match('^bootdiagnostics-', container.name):
+        if container.name.startswith("bootdiagnostics-"):
```
asmorodskyi, line 138: for me it is the same , why `startswith` is better ?

ocw/lib/aks.py
```diff
@@ -15,33 +12,12 @@ class AKS(Provider):
     def __new__(cls, vault_namespace):
         if vault_namespace not in AKS.__instances:
             AKS.__instances[vault_namespace] = self = object.__new__(cls)
-            self.__container_mgmt_client = None
```
asmorodskyi, line 18: @ilausuch  please check this out . WDYT ?

asmorodskyi (approved):
LGTM but I would like to get confirmation from @ilausuch  about code cleanup in aks

ocw/lib/azure.py
ricardobranco777: > for me it is the same , why `startswith` is better ?

It's more Pythonic using a string method instead of the `re` module which makes it Perlish. ;)

grisu48 (commented):
LGTM, except for the changes in `ocw/lib/aks.py`, where I cannot comment on. The rest :+1: 

I do like that we are getting rid of some of the regex-specifics and replace them with better comprehensible python function, which reduces the mental effort required to read the code :+1:

ocw/lib/aks.py
ilausuch: I tested locally, and seems that works without this. Good catch

ilausuch (approved):

grisu48 (approved):

PR #275, opened 2023-07-03
Fix kubernetes package version

ilausuch:
The current package kubernetes1.23 doesn't exist anymore, therefore the build of containers fail. Next available version is 1.24

### DIFF
--- a/Dockerfile_k8s
+++ b/Dockerfile_k8s
@@ -1,6 +1,6 @@
 FROM registry.suse.com/bci/python:3.10
 
-RUN zypper -n in gcc tar gzip kubernetes1.23-client aws-cli azure-cli && zypper clean && rm -rf /var/cache
+RUN zypper -n in gcc tar gzip kubernetes1.24-client aws-cli azure-cli && zypper clean && rm -rf /var/cache
 
 # Google cli installation
 RUN curl -sf https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-415.0.0-linux-x86_64.tar.gz | tar -zxf - -C /opt \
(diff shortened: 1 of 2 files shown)


ricardobranco777 (approved):
LGTM

grisu48 (approved):
LGTM, although a PR description would be very nice to have :slightly_smiling_face:

asmorodskyi:
how we test that this work ? keeping in mind that service currently is silent it is very important question

ilausuch:
I updated the commit too

ilausuch:
> how we test that this work ? keeping in mind that service currently is silent it is very important question

We have in our workflows a build matrix with pcw and k8s. However if the deprecation of version 1.23 is recent we didn't detect

asmorodskyi:
> > how we test that this work ? keeping in mind that service currently is silent it is very important question
> 
> We have in our github workflows a build matrix with pcw and k8s. However if the deprecation of version 1.23 is recent we didn't detect

this is about building container , what about execution of container ? what if 1.24 requires some new parameter and throw exception if it is not defined ?

ilausuch:
> > > how we test that this work ? keeping in mind that service currently is silent it is very important question
> > 
> > 
> > We have in our github workflows a build matrix with pcw and k8s. However if the deprecation of version 1.23 is recent we didn't detect
> 
> this is about building container , what about execution of container ? what if 1.24 requires some new parameter and throw exception if it is not defined ?

It's tested locally but we have this ticket https://progress.opensuse.org/issues/130144 we can create a ticket for testing this

ilausuch:
however, we should update our k8s servers ensuring that are compatible with the last version of kubernetes client in our packages. We did on AKS recently

PR #276, opened 2023-07-03
Increase the verbose for cleanup_k8s

ilausuch:
https://progress.opensuse.org/issues/131522

### DIFF
--- a/cleanup_k8s.py
+++ b/cleanup_k8s.py
@@ -1,13 +1,19 @@
+import os
 import logging
 from ocw.lib.gke import GKE
 from ocw.lib.eks import EKS
 from ocw.lib.aks import AKS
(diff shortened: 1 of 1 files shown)


ricardobranco777 (approved):
LGTM

cleanup_k8s.py
```diff
@@ -8,6 +8,7 @@
 
 def main():
     logger = logging.getLogger(__name__)
+    logging.basicConfig(level=logging.INFO)
```
asmorodskyi, line 11: https://github.com/SUSE/pcw/blob/master/webui/settings.py#L145

please use same approach

ilausuch: done

PR #277, opened 2023-07-03
Minor fixes

ricardobranco777:
Minor fixes:
- Use pylint on more files and fix issues that appeared.
- Avoid `dangerous-default-value`.
- Avoid `re` where unneeded.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -6,7 +6,6 @@ disable=
  C0206, # Consider iterating with .items()
  C0415, # Import outside toplevel
  E1101, # no-member
- R0904, # Too many public methods
  R1702, # Too many nested blocks
  W0237, # arguments-renamed
  W0603, # Using the global statement
--- a/tests/kubernetes.py
+++ b/tests/kubernetes.py
@@ -7,7 +7,9 @@ class MockedKubernetesConfig():
 
 
 class MockedKubernetesClient():
-    def __init__(self, jobs=[]):
+    def __init__(self, jobs=None):
+        if jobs is None:
+            jobs = []
         self.jobs = jobs
         self.deleted_jobs = []
 
--- a/webui/PCWConfig.py
+++ b/webui/PCWConfig.py
@@ -44,7 +44,9 @@ class ConfigFile:
                 return default
         return config_pointer
 
-    def getList(self, config_path: str, default: list = []) -> list:
+    def getList(self, config_path: str, default: list = None) -> list:
+        if default is None:
+            default = []
         return [i.strip() for i in self.get(config_path, ','.join(default)).split(',')]
 
 
@@ -83,7 +85,7 @@ class PCWConfig():
     def get_namespaces_for(feature: str) -> list:
         if PCWConfig.has(feature):
             return ConfigFile().getList(f'{feature}/namespaces', ConfigFile().getList('default/namespaces'))
-        return list()
+        return []
 
     @staticmethod
     def get_providers_for(feature: str, namespace: str):
@@ -114,7 +116,7 @@ class PCWConfig():
     @staticmethod
     def getBoolean(config_path: str, namespace: str = None, default=False) -> bool:
         if namespace:
-            (feature, property) = config_path.split('/')
+            feature, property = config_path.split('/')
             setting = f'{feature}.namespace.{namespace}/{property}'
             if PCWConfig.has(setting):
                 value = ConfigFile().get(setting)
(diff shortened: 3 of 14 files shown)


webui/PCWConfig.py
```diff
@@ -83,7 +85,7 @@ def get_feature_property(feature: str, property: str, namespace: str = None):
     def get_namespaces_for(feature: str) -> list:
         if PCWConfig.has(feature):
             return ConfigFile().getList(f'{feature}/namespaces', ConfigFile().getList('default/namespaces'))
-        return list()
+        return []
```
asmorodskyi, line 88: I remember that I was corrected by some linter to actually use `list()` :thinking:

pylintrc
```diff
@@ -7,7 +7,9 @@ disable=
  C0415, # Import outside toplevel
  E1101, # no-member
  R0904, # Too many public methods
+ R0903, # Too few public methods
```
asmorodskyi, line 10: which class produces this ? does it make sense to block it on global level ? maybe we will put this into certain class ?

pylintrc
```diff
@@ -7,7 +7,9 @@ disable=
...
  R1702, # Too many nested blocks
+ R0901, # Too many ancestors
```
asmorodskyi, line 12: is it about Provider  ? maybe we should just put ignorance there ?

pylintrc
ricardobranco777: In this case I prefer to put it here since we already have its complement (Too many public methods) above.  Also, both are what I consider "stupid" warnings by Pylint since it's common to create classes with too many and too few public methods.

pylintrc
ricardobranco777: > is it about Provider ? maybe we should just put ignorance there ?

FilteredSingleTableView in ocw/views.py

pylintrc
asmorodskyi: 1. R0904 is here because when I was introducing pylint pipeline I have too many things to do so I needed to accept certain warnings to be able to merge whole idea . So as all other items R0904 is candidate to removal together with fix of the code which causing this error. 
2.  I can not agree that R0903 and R0904 are stupid . R0903 signaling you - "maybe your class can be replaced by dictionary ? "  and R0904 signaling you - "maybe you class will looks better as 2 separate classes ? ". Of course both questions may be ignored in some certain cases but I think blocking linter ability to ask them globally does not sounds good

webui/PCWConfig.py
asmorodskyi: you just resolved w/o putting answer :)

ricardobranco777: > you just resolved w/o putting answer :)

I think in this case we have to agree with the linter, even if it contradicts a former version of itself. :)

asmorodskyi: ah ok I get the point now . yeah "Rule 1: Linter always right Rule 2: when linter not right see Rule 1" :D

pylintrc
ricardobranco777: > 2. I can not agree that R0903 and R0904 are stupid . R0903 signaling you - "maybe your class can be replaced by dictionary ? "  and R0904 signaling you - "maybe you class will looks better as 2 separate classes ? ". Of course both questions may be ignored in some certain cases but I think blocking linter ability to ask them globally does not sounds good

The issue with Django is that subclassing with few methods is unavoidable, so I'm ignoring them at the class level now.

Removed both from pylintrc.  EC2 will need a refactor to decrease the number of methods from 23 to <= 20 but it's outside the scope of this PR.

grisu48 (approved):

tests/kubernetes.py
```diff
@@ -7,7 +7,9 @@ def load_kube_config(self, *args, **kwargs):
 
 
 class MockedKubernetesClient():
-    def __init__(self, jobs=[]):
+    def __init__(self, jobs=None):
```
asmorodskyi, line 10: it is basically the same but more lines of code. What is benefit ?

ricardobranco777: > it is basically the same but more lines of code. What is benefit ?

Using mutable objects such as lists and dicts for default values in function definition is dangerous and could have side-effects.  It's common Python gotcha as documented [here](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).  Consider the following code:

```
def f(array=[]):
    array.append(1)
    return array

def f2(array=None):
    if array is None:
        array = []
    array.append(1)
    return array

print(f())
print(f())

print(f2())
print(f2())
```

It prints:
```
[1]
[1, 1]
[1]
[1]
```

PR #278, opened 2023-07-12
Discover cancelled jobs to remove resources earlier than expired TTL

ricardobranco777:
Currently, if we manually cancel VR's the instances may remain active until pcw cleans them up when the TTL expires.

This PR:
- Uses the openQA client to discover those cancelled jobs, which we get from the instance's tags.
- Adds `get_url()` to guess the protocol used by the openQA server if no scheme is present.
- Adds unit tests to `ocw.models`
- Removes redundant `browser-actions/setup-geckodriver` because of rate-limits.

TODO:
- Test it.

### DIFF
--- a/.github/workflows/pcw.yml
+++ b/.github/workflows/pcw.yml
@@ -14,7 +14,6 @@ jobs:
         python-version: ['3.10']
     steps:
       - uses: actions/checkout@v3
-      - uses: browser-actions/setup-geckodriver@latest
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
--- a/ocw/lib/openqa.py
+++ b/ocw/lib/openqa.py
@@ -1,10 +1,17 @@
+import os
 from urllib.parse import urlparse
 from cachetools import cached
-import requests
 from requests.exceptions import RequestException
-from openqa_client.client import OpenQA_Client
+import openqa_client.client
 from openqa_client.const import JOB_STATE_CANCELLED
 from openqa_client.exceptions import OpenQAClientError
+
+
+if not bool(os.environ.get('REQUESTS_CA_BUNDLE')):
+    # Disable urllib3 warnings for InsecureRequestWarning only on openqa_client.client
+    openqa_client.client.requests.packages.urllib3.disable_warnings(
+        openqa_client.client.requests.packages.urllib3.exceptions.InsecureRequestWarning
+    )
 
 
 @cached(cache={})
@@ -16,7 +23,11 @@ def get_url(server):
     for scheme in ("https", "http"):
         try:
             url = f"{scheme}://{server}"
-            got = requests.head(url, timeout=30)
+            got = openqa_client.client.requests.head(
+                url,
+                timeout=5,
+                verify=bool(os.environ.get('REQUESTS_CA_BUNDLE')),
+            )
             got.raise_for_status()
             return url
         except RequestException:
@@ -36,7 +47,8 @@ class OpenQA:
 
     def __init__(self, **kwargs):
         kwargs.pop("server")
-        self.__client = OpenQA_Client(server=self.server, **kwargs)
+        self.__client = openqa_client.client.OpenQA_Client(server=self.server, **kwargs)
+        self.__client.session.verify = bool(os.environ.get('REQUESTS_CA_BUNDLE'))
 
     def is_cancelled(self, job_id: str) -> bool:
         if not job_id.isdigit():
--- a/ocw/lib/openqa.py
+++ b/ocw/lib/openqa.py
@@ -1,0 +1,50 @@
+from urllib.parse import urlparse
+from cachetools import cached
+import requests
+from requests.exceptions import RequestException
+from openqa_client.client import OpenQA_Client
+from openqa_client.const import JOB_STATE_CANCELLED
+from openqa_client.exceptions import OpenQAClientError
+
+
+@cached(cache={})
+def get_url(server):
+    if server:
+        server = server.rstrip('/').replace("_", ".")
+    if urlparse(server).scheme != "":
+        return server
+    for scheme in ("https", "http"):
+        try:
+            url = f"{scheme}://{server}"
+            got = requests.head(url, timeout=30)
+            got.raise_for_status()
+            return url
+        except RequestException:
+            pass
+    raise OpenQAClientError(f"Could not connect to server {server}")
+
+
+class OpenQA:
+    __servers = {}
+
+    def __new__(cls, **kwargs):
+        server = urlparse(get_url(kwargs["server"])).netloc
+        if server not in cls.__servers:
+            cls.__servers[server] = self = super().__new__(cls)
+            self.server = server
+        return cls.__servers[server]
+
+    def __init__(self, **kwargs):
+        kwargs.pop("server")
+        self.__client = OpenQA_Client(server=self.server, **kwargs)
+
+    def is_cancelled(self, job_id: str) -> bool:
+        if not job_id.isdigit():
+            raise ValueError(f"job must be a number: {job_id}")
+        try:
+            status = self.__client.openqa_request('GET', f'jobs/{job_id}')
+            # Return true if job is either cancelled or done
+            return status['job']['state'] == JOB_STATE_CANCELLED
+        except (TypeError, KeyError, OpenQAClientError):
+            pass
+        return False
--- a/ocw/models.py
+++ b/ocw/models.py
@@ -1,19 +1,23 @@
 import json
+import logging
 from datetime import datetime, timedelta, timezone
 from django.db import models
-from webui.PCWConfig import PCWConfig
 from .enums import ProviderChoice, StateChoice
+from .lib.openqa import OpenQA, get_url, OpenQAClientError
 
 
 def format_seconds(seconds):
+    if not seconds:
+        return "0s"
     days, remainder = divmod(seconds, 60*60*24)
     hours, remainder = divmod(remainder, 60*60)
     minutes, seconds = divmod(remainder, 60)
-    if days > 0:
-        return f'{days:.0f}d{hours:.0f}h{minutes:.0f}m'
-    if hours > 0:
-        return f'{hours:.0f}h{minutes:.0f}m'
-    return f'{minutes:.0f}m'
+    return "".join([
+        f"{days:.0f}d" if days > 0 else "",
+        f"{hours:.0f}h" if hours > 0 else "",
+        f"{minutes:.0f}m" if minutes > 0 else "",
+        f"{seconds:.0f}s" if seconds > 0 else "",
+    ])
 
 
 class Instance(models.Model):
@@ -32,11 +36,14 @@ class Instance(models.Model):
     ignore = models.BooleanField(default=False)
     TAG_IGNORE = 'pcw_ignore'
 
-    def age_formated(self):
+    def age_formatted(self):
         return format_seconds(self.age.total_seconds())
 
-    def ttl_formated(self):
-        return format_seconds(self.ttl.total_seconds()) if (self.ttl) else ""
+    def ttl_formatted(self):
+        return format_seconds(self.ttl.total_seconds())
+
+    def ttl_expired(self):
+        return self.age.total_seconds() > self.ttl.total_seconds()
 
     def all_time_fields(self):
         all_time_pattern = "(age={}, first_seen={}, last_seen={}, ttl={})"
@@ -46,7 +53,7 @@ class Instance(models.Model):
             first_fmt = self.first_seen.strftime('%Y-%m-%d %H:%M')
         if self.last_seen:
             last_fmt = self.last_seen.strftime('%Y-%m-%d %H:%M')
-        return all_time_pattern.format(self.age_formated(), first_fmt, last_fmt, self.ttl_formated())
+        return all_time_pattern.format(self.age_formatted(), first_fmt, last_fmt, self.ttl_formatted())
 
     def set_alive(self):
         self.last_seen = datetime.now(tz=timezone.utc)
@@ -58,6 +65,20 @@ class Instance(models.Model):
 
     def get_type(self):
         return self.cspinfo.type
+
+    def is_cancelled(self) -> bool:
+        job = self.cspinfo.get_tag("openqa_var_job_id")
+        server = self.cspinfo.get_tag("openqa_var_server")
+        if job and server:
+            try:
+                return OpenQA(
+                    server=server,
+                    retries=1,  # default is 5
+                    wait=5,     # default is 10
+                ).is_cancelled(job)
+            except OpenQAClientError as exc:
+                logging.warning("%s: %s", server, exc)
+        return False
 
     class Meta:  # pylint: disable=too-few-public-methods
         unique_together = (('provider', 'instance_id', 'vault_namespace'),)
@@ -73,10 +94,18 @@ class CspInfo(models.Model):
     type = models.CharField(max_length=200)
 
     def get_openqa_job_link(self):
-        if self.get_tag('openqa_created_by') == 'openqa-suse-de' and self.get_tag('openqa_var_JOB_ID') is not None:
-            url = f"{PCWConfig.get_feature_property('webui', 'openqa_url')}/t{self.get_tag('openqa_var_JOB_ID')}"
-            title = self.get_tag('openqa_var_NAME', '')
-            return {'url': url, 'title': title}
+        server = self.get_tag('openqa_var_server')
+        if server:
+            try:
+                server = get_url(server)
+            except OpenQAClientError as exc:
+                logging.warning("%s: %s", server, exc)
+                return None
+            job_id = self.get_tag('openqa_var_job_id')
+            if job_id:
+                url = f"{server}/t{job_id}"
+                title = self.get_tag('openqa_var_name', '')
+                return {'url': url, 'title': title}
         return None
 
     def get_tag(self, tag_name, default_value=None):
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ azure-mgmt-storage==20.0.0
 azure-mgmt-resource==21.1.0
 azure-storage-blob==12.13.0
 azure-identity==1.10.0
+cachetools
 msrestazure==0.6.4
 uwsgi==2.0.21
 requests==2.31.0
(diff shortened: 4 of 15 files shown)


ricardobranco777:
> generally I like the idea but we need to take into consideration several things :
> 
> 1. There a multiply openQA instances which can create VM in the cloud and currently we posting only job_id ( AFAIK ) . This logic certainly need to take this in consideration

Can you please elaborate on this?

> 2. In openQA we have special variable which keep VM running after failed job we should obey it

This PR currently obeys the "pcw-ignore" tag.

ocw/lib/provider.py
```diff
@@ -35,6 +38,31 @@ def get_data(self, name=None) -> str:
...
+        job = tags.get("openqa_var_JOB_ID")
+        if job is None:
+            # Parse "11555642 openqa_created_by=openqa.suse.de/t11555642 openqa_ttl=11100"
+            try:
+                job = tags.get("openqa-cli-test-tag").split()[0]
+            except AttributeError:
+                pass
+        if job is not None:
+            return OpenQA(
+                server=OPENQA_URL,
```
mpagot, line 61: Does not this create an hard dependency between the generic PCW code and a setup referring to OSD?

What happens if the job id referred in the `openqa_var_JOB_ID` is not found on OSD instance ? It could happens if an ID does not refer to a job on OSD but on some other instance.

mpagot: ... or just means that it is a feature of PCW that only do something for a specific subset of resources:
- resources with a specific tags ...
- ... that refer to a job id that exist on OSD ...
- ... and  is in status cancelled.

The feature remain inactive for anything else. (resource without that tag, resources from job on other openQA instances, resources from job not cancelled but maybe failed).

pdostal: From my POV:
1) Job not found in openQA should be ignored, not deleted.
2) Jobs without `openqa_created_by` and `openqa_ttl` tags should be also ignored.
3) Not only OSD should be supported but also other instances.

ricardobranco777: > Does not this create an hard dependency between the generic PCW code and a setup referring to OSD?

This is already the case within `webui`.  PCW is not generic as this is our tool for cleanup.

The `OpenQA` class introduced here supports multiple servers, though.  It could parse the URL found in `openqa_created_by`.  I know that you're using a dedicated machine for openQA jobs.  That's why I wanted your input.

> What happens if the job id referred in the `openqa_var_JOB_ID` is not found on OSD instance ? It could happens if an ID does not refer to a job on OSD but on some other instance.

The job _should_ have a known tag.

asmorodskyi:
> 
> Can you please elaborate on this?
> 
> I found these cases:
> 
> ```
> {
>   "openqa-cli-test-tag": "11556015",
>   "openqa_created_by": "openqa.suse.de/t11556015",
>   "openqa_ttl": "11100"
> }
> ```

if you search better you might find something like this : 

```
{
  "openqa_created_by": "openqa-suse-de",
  "openqa_created_date": "2023-07-12T09:48:36Z",
  "openqa_created_id": "e5bd4a958ea0dfe6",
  "openqa_ttl": "11100",
 **"openqa_var_JOB_ID": "1155",**
  "openqa_var_NAME": "11554458-sle-15-SP5-EC2-Updates-x86_64-Build20230712-1-publiccloud_img_proof@64bit"
}
```

if you try to run this against https://openqa.suse.de  you will get 404 because there is no such job ID. And the reason that this VM was actually created by http://autobot.qe.nue2.suse.org .  Ofc. with your current code it is not that big problem PCW will simply delete it in normal flow. But still I think if we want to introduce such logic we need to take it into account. To achieve this we also need to tweak test code. Tags suppose to look like this :

```
{
  "openqa_created_by": "openqa-suse-de",
  "openqa_created_date": "2023-07-12T09:48:36Z",
  "openqa_created_id": "e5bd4a958ea0dfe6",
  "openqa_ttl": "11100",
 "openqa_var_JOB_ID": "1155",
 "openqa_var_SERVER": "autobot.qe.nue2.suse.org",
  "openqa_var_NAME": "11554458-sle-15-SP5-EC2-Updates-x86_64-Build20230712-1-publiccloud_img_proof@64bit"
}
```


> 
> > 2. In openQA we have special variable which keep VM running after failed job we should obey it
> 
> This PR currently obeys the "pcw-ignore" tag.

No I was not speaking about "pcw-ignore" , I was speaking about `PUBLIC_CLOUD_NO_CLEANUP_ON_FAILURE` openQA variable . This variable is blocking terraform delete so VM will run after test failure for defined TTL . You logic will break this model

ocw/lib/provider.py
asmorodskyi: > This is already the case within `webui`. PCW is not generic as this is our tool for cleanup.

I think we need to change this . see my comment above . we need to drop this dependency to **certain** openQA instance . Each set of tags should contain not only job id  but also server on which this job was running. 

> > What happens if the job id referred in the `openqa_var_JOB_ID` is not found on OSD instance ? It could happens if an ID does not refer to a job on OSD but on some other instance.
> 
> The job _should_ have a known tag.

I think Michele meant here is something what I said originally - we should cover the case that VM was created NOT by OSD.

ricardobranco777: > I think we need to change this . see my comment above . we need to drop this dependency to **certain** openQA instance . 

Will see how feasible it is to implement this change in this PR>

> Each set of tags should contain not only job id but also server on which this job was running.

Fully agree, and we should not have non-URL tag values with `openqa-suse-de`.

> I think Michele meant here is something what I said originally - we should cover the case that VM was created NOT by OSD.

Yes.

ocw/lib/azure.py
```diff
@@ -150,7 +150,7 @@ def cleanup_blob_containers(self) -> None:
                 self.log_dbg(f'Found container {container.name}')
                 container_blobs = self.container_client(container.name).list_blobs()
                 for blob in container_blobs:
-                    if self.is_outdated(blob.last_modified):
+                    if self.is_outdated(blob.last_modified) or self.is_cancelled(blob):
```
asmorodskyi, line 153: I would not propagate this on disks , images and everything which goes into cleanup_all logic . this entities has different lifecycle and not tightly coupled to certain job . also this resource type NOT always have needed tags ( actually better to say in most cases they do NOT have them ) 

Also please check this [URL ](https://openqa.suse.de/tests/11560368#dependencies)  to see why it is not gonna work

asmorodskyi (changes requested):
we need to drop this logic from everything outside ocw.lib.db.auto_delete_instances()

ocw/lib/ec2.py
```diff
@@ -70,7 +77,7 @@ def cleanup_snapshots(self, valid_period_days: float) -> None:
             response = self.ec2_client(region).describe_snapshots(OwnerIds=['self'])
             self.log_dbg(f"Found {len(response['Snapshots'])} snapshots in {region}")
             for snapshot in response['Snapshots']:
-                if EC2.is_outdated(snapshot['StartTime'], valid_period_days):
+                if self.is_outdated(snapshot['StartTime'], valid_period_days) or self.is_cancelled(snapshot):
```
asmorodskyi, line 80: is_outdated is static method , while what you have written will work I prefer to have `EC2.is_outdated` this way we underline that we calling method which has nothing to do with certain object instance and does not operate any runtime data

ocw/lib/openqa.py
```diff
@@ -0,0 +1,79 @@
+from urllib.parse import urlparse
+import requests
+from requests.exceptions import RequestException
+
+# pylint: disable=import-error
```
asmorodskyi, line 5: why do we have import error here ?

ocw/lib/openqa.py
```diff
@@ -0,0 +1,79 @@
...
+    raise OpenQAClientError(f"Could not connect to server {server}")
+
+
+class OpenQA:
+    __servers = {}
+
+    def __new__(cls, **kwargs):
+        if kwargs["server"] not in cls.__servers:
+            server = get_url(kwargs["server"])
+            # Use same object for "https://openqa.suse.de", "openqa.suse.de" & "https://openqa.suse.de/"
```
asmorodskyi, line 34: I am not sure why we need this complication . `get_url` function **always** returning url in same format so how do we expect to get something else except `https://openqa.suse.de` ?

asmorodskyi: I mean that no matter what was used for init `OpenQA` object :
```
server1 = openQA("https://openqa.suse.de")
server2=openQA("openqa.suse.de")
server3=openQA("https://openqa.suse.de/")
```

`__servers` array suppose to get exactly same , than later we don't need to play a guess game here

ocw/lib/openqa.py
```diff
@@ -0,0 +1,79 @@
...
+            self.server = server
+        return cls.__servers[kwargs["server"]]
+
+    def __init__(self, **kwargs):
+        kwargs.pop("server")
+        self.__client = OpenQA_Client(server=self.server, **kwargs)
+
+    def is_cancelled(self, job: str) -> bool:
+        if not job.isdigit():
+            raise ValueError(f"Invalid job: {job}")
```
asmorodskyi, line 50:
```suggestion
            raise ValueError(f"Job id must be a digit: {job}")
```

ocw/lib/openqa.py
```diff
@@ -0,0 +1,79 @@
...
+            for key in (server, urlparse(server).netloc, kwargs["server"]):
+                cls.__servers[key] = self
+            self.server = server
+        return cls.__servers[kwargs["server"]]
+
+    def __init__(self, **kwargs):
+        kwargs.pop("server")
+        self.__client = OpenQA_Client(server=self.server, **kwargs)
+
+    def is_cancelled(self, job: str) -> bool:
```
asmorodskyi, line 48:
```suggestion
    def is_cancelled(self, job_id: str) -> bool:
```

ocw/lib/openqa.py
```diff
@@ -0,0 +1,79 @@
...
+        return False
+
+
+def is_cancelled(tags) -> bool:
+    if TAG_IGNORE in tags:
+        return False
+    job = tags.get("openqa_var_JOB_ID")
+    if job is None:
+        # Parse "11555642 openqa_created_by=openqa.suse.de/t11555642 openqa_ttl=11100"
+        for tag in ('openqa-cli-test-tag', 'openqa-cli-test-label'):
```
asmorodskyi, line 66: what we trying to do here ? Let's not try to be too smart . If `openqa_var_JOB_ID` does not presented we simply should return False

ocw/lib/openqa.py
```diff
@@ -0,0 +1,79 @@
...
+            try:
+                job = tags.get(tag).split()[0]
+                break
+            except AttributeError:
+                continue
+    if job is None:
+        return False
+    server = tags.get("openqa_var_SERVER", "https://openqa.suse.de")
```
asmorodskyi, line 74: Again this whole flow should work ONLY if all pieces are presented . If something is missing we just return False . And fallback to our usual flow 
```suggestion
    server = tags.get("openqa_var_SERVER")
```

ocw/models.py
```diff
@@ -73,10 +72,19 @@ class CspInfo(models.Model):
...
-            url = f"{PCWConfig.get_feature_property('webui', 'openqa_url')}/t{self.get_tag('openqa_var_JOB_ID')}"
+        server = self.get_tag('openqa_var_SERVER')
+        job_id = self.get_tag('openqa_var_JOB_ID')
+        if server and job_id:
+            url = f"{server}/t{job_id}"
             title = self.get_tag('openqa_var_NAME', '')
             return {'url': url, 'title': title}
+        created_by = self.get_tag('created_by')
+        if created_by is not None:
+            if "/t" in created_by:
```
asmorodskyi, line 83: why ? which flow we supporting here ?

ocw/lib/openqa.py
ricardobranco777: > I am not sure why we need this complication . `get_url` function **always** returning url in same format so how do we expect to get something else except `https://openqa.suse.de` ?

I want to avoid calling `get_url()` everytime `OpenQA` is called with the hostname without the scheme because `get_url()`  tries to guess the protocol.  So either we prepare for it here or use a `lru_cache`  decorator in `get_url`.

ocw/lib/openqa.py
ricardobranco777: > what we trying to do here ? Let's not try to be too smart . If `openqa_var_JOB_ID` does not presented we simply should return False

The cli-test jobs don' t use `openqa_var_JOB_ID` at all.

ocw/models.py
ricardobranco777: > why ? which flow we supporting here ?

The `$provider_cli`  tests.

ocw/lib/openqa.py
ricardobranco777: I removed support for such tests with the intention of fixing it in our test code.

ocw/models.py
asmorodskyi: but `created_by` should not contain test url in any case so why we have test which expects something like this ? maybe better tweak/drop such test instead of aligning prod code with it ?

ricardobranco777: > but `created_by` should not contain test url in any case so why we have test which expects something like this ? maybe better tweak/drop such test instead of aligning prod code with it ?

Then I would do that in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17420

It does contain URL's:

```
$ grep -r created_by tests/publiccloud/
tests/publiccloud/aws_cli.pm:    my $created_by = "$openqa_url/t$job_id";
tests/publiccloud/aws_cli.pm:    my $tag = "{Key=openqa-cli-test-tag,Value=$job_id},{Key=openqa_created_by,Value=$created_by},{Key=openqa_ttl,Value=$openqa_ttl}";
tests/publiccloud/azure_cli.pm:    my $created_by = "$openqa_url/t$job_id";
tests/publiccloud/azure_cli.pm:    my $tags = "openqa-cli-test-tag=$job_id openqa_created_by=$created_by openqa_ttl=$openqa_ttl";
tests/publiccloud/azure_more_cli.pm:    my $created_by = "$openqa_url/t$job_id";
tests/publiccloud/azure_more_cli.pm:    my $tags = "openqa-cli-test-tag=$job_id openqa_created_by=$created_by openqa_ttl=$openqa_ttl";
tests/publiccloud/google_cli.pm:    my $labels = "openqa-cli-test-label=$job_id,openqa_created_by=$openqa_hostname,openqa_ttl=$openqa_ttl"
```

ocw/lib/openqa.py
asmorodskyi: ok I see , make sense than

asmorodskyi (approved):
LGTM

asmorodskyi:
I am confident with code now . Good job Ricardo ! 
But because you touching very sensitive piece of code ( actual logic deciding which VMs to delete ) we need to do some good testing ... I am not sure what actually needs to be done ... need to think about this

ocw/models.py
```diff
@@ -59,6 +65,19 @@ def set_alive(self):
     def get_type(self):
         return self.cspinfo.type
 
+    def is_cancelled(self) -> bool:
```
asmorodskyi, line 68: nice ! I like it !

pdostal (approved):

.github/workflows/pcw.yml
```diff
@@ -14,7 +14,6 @@ jobs:
         python-version: ['3.10']
     steps:
       - uses: actions/checkout@v3
-      - uses: browser-actions/setup-geckodriver@latest
```
asmorodskyi, line 17: why ?

ocw/models.py
```diff
@@ -73,8 +93,13 @@ class CspInfo(models.Model):
     type = models.CharField(max_length=200)
 
     def get_openqa_job_link(self):
-        if self.get_tag('openqa_created_by') == 'openqa-suse-de' and self.get_tag('openqa_var_JOB_ID') is not None:
-            url = f"{PCWConfig.get_feature_property('webui', 'openqa_url')}/t{self.get_tag('openqa_var_JOB_ID')}"
+        try:
+            server = get_url(self.get_tag('openqa_var_server'))
```
asmorodskyi, line 97: what if get_tag will return None here ?

requirements.txt
```diff
@@ -4,6 +4,7 @@ azure-mgmt-storage==20.0.0
 azure-mgmt-resource==21.1.0
 azure-storage-blob==12.13.0
 azure-identity==1.10.0
+cachetools
```
asmorodskyi, line 7: where we using this ?

.github/workflows/pcw.yml
ricardobranco777: > why ?

It's redundant because geckodriver it's already in the VM used by GitHub Actions, and we risk running into rate-limits when pulling this action like it happened to me today.  See here:

https://github.com/SUSE/pcw/actions/runs/5647942036/job/15299079594

> Run browser-actions/setup-geckodriver@latest
> with:
> Error: API rate limit exceeded for 20.57.79.241. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

ocw/lib/openqa.py
```diff
@@ -0,0 +1,62 @@
...
+            got = openqa_client.client.requests.head(
+                url,
+                timeout=5,
+                verify=bool(os.environ.get('REQUESTS_CA_BUNDLE')),
+            )
+            got.raise_for_status()
+            return url
+        except RequestException:
+            pass
+    raise OpenQAClientError(f"Could not connect to server {server}")
```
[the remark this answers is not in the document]
ricardobranco777: > can you elaborate why we need to throw error here ? later everywhere in the code we ignoring/muting this exception . So maybe better just not to throw it ?

It's better to throw it and catch it than returning a null string and comparing, IMO.  I will add a warning print in the places where we catch it to help debugging.

ocw/models.py
ricardobranco777: > what if get_tag will return None here ?

It won't pass the `if server and job_id` check later.

requirements.txt
ricardobranco777: > where we using this ?

We use the `cached` decorator in `ocw.lib.openqa.get_url()` to cache the results.

ocw/models.py
ricardobranco777: Fixed.

.github/workflows/pcw.yml
asmorodskyi: :+1:

ocw/lib/openqa.py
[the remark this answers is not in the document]
asmorodskyi: ok

PR #280, opened 2023-07-25
Add support for cleaning up Azure Image Galleries

ricardobranco777:
This PR:
- Adds support for cleaning up Azure Image Galleries.

TODO:
- Refactor Azure test code.

### DIFF
--- a/ocw/lib/azure.py
+++ b/ocw/lib/azure.py
@@ -7,6 +7,7 @@ from azure.mgmt.storage import StorageManagementClient
 from azure.storage.blob import BlobServiceClient
 from azure.core.exceptions import ResourceNotFoundError
 from msrest.exceptions import AuthenticationError
+from dateutil.parser import parse
 from webui.PCWConfig import PCWConfig
 from .provider import Provider
 from ..models import Instance
@@ -19,6 +20,7 @@ class Azure(Provider):
         super().__init__(namespace)
         self.__resource_group = PCWConfig.get_feature_property('cleanup', 'azure-storage-resourcegroup', namespace)
         self.check_credentials()
+        self.__gallery = PCWConfig.get_feature_property('cleanup', 'azure-gallery-name', namespace)
 
     def __new__(cls, vault_namespace: str) -> 'Azure':
         if vault_namespace not in Azure.__instances:
@@ -96,6 +98,9 @@ class Azure(Provider):
             return ', '.join(type_set)
         return "N/A"
 
+    def get_resource_properties(self, resource_id):
+        return self.resource_mgmt_client().resources.get_by_id(resource_id, api_version="2023-07-03").properties
+
     def list_resource_groups(self) -> list:
         return list(self.resource_mgmt_client().resource_groups.list())
 
@@ -106,22 +111,21 @@ class Azure(Provider):
             self.log_info(f"Deleting of resource group {resource_id}")
             self.resource_mgmt_client().resource_groups.begin_delete(resource_id)
 
-    def list_images_by_resource_group(self, resource_group):
-        return self.list_by_resource_group(resource_group,
-                                           filters="resourceType eq 'Microsoft.Compute/images'")
+    def list_images(self):
+        return self.list_resource(filters="resourceType eq 'Microsoft.Compute/images'")
 
-    def list_disks_by_resource_group(self, resource_group):
-        return self.list_by_resource_group(resource_group,
-                                           filters="resourceType eq 'Microsoft.Compute/disks'")
+    def list_disks(self):
+        return self.list_resource(filters="resourceType eq 'Microsoft.Compute/disks'")
 
-    def list_by_resource_group(self, resource_group, filters=None) -> list:
+    def list_resource(self, filters=None) -> list:
         return list(self.resource_mgmt_client().resources.list_by_resource_group(
-            resource_group, filter=filters, expand="changedTime"))
+            self.__resource_group, filter=filters, expand="changedTime"))
 
     def cleanup_all(self) -> None:
         self.log_info("Call cleanup_all")
-        self.cleanup_images_from_rg()
-        self.cleanup_disks_from_rg()
+        self.cleanup_images()
+        self.cleanup_gallery_img_versions()
+        self.cleanup_disks()
         self.cleanup_blob_containers()
 
     @staticmethod
@@ -157,9 +161,9 @@ class Azure(Provider):
                             self.log_info(f"Deleting blob {blob.name}")
                             self.container_client(container.name).delete_blob(blob.name, delete_snapshots="include")
 
-    def cleanup_images_from_rg(self) -> None:
-        self.log_dbg("Call cleanup_images_from_rg")
-        for item in self.list_images_by_resource_group(self.__resource_group):
+    def cleanup_images(self) -> None:
+        self.log_dbg("Call cleanup_images")
+        for item in self.list_images():
             if self.is_outdated(item.changed_time):
                 if self.dry_run:
                     self.log_info(f"Deletion of image {item.name} skipped due to dry run mode")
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -5,12 +5,10 @@ ENV UWSGI_HTTP_AUTO_CHUNKED=1 UWSGI_HTTP_KEEPALIVE=1 UWSGI_LAZY_APPS=1 UWSGI_WSG
...
 
-# We do the whole installation and configuration in one layer:
+RUN zypper -n in gcc libffi-devel && zypper clean && rm -rf /var/cache
+
 COPY requirements.txt /pcw/
-# * Install system requirements
-# * Install pip requirements
-# * Empty system cache to conserve some space
-RUN zypper -n in gcc libffi-devel && pip install --no-cache-dir -r /pcw/requirements.txt && zypper clean && rm -rf /var/cache
+RUN pip install --no-cache-dir -r /pcw/requirements.txt
```
asmorodskyi, line 11: this will dramatically increase overall weight of container because you basically adding extra layer where we had one . What the benefit of that ?

ricardobranco777: > this will dramatically increase overall weight of container because you basically adding extra layer where we had one . What the benefit of that ?

Size increased by only 90MB but I suspect it came mostly from updated dependencies.

The benefit is that I don't have to wait for OS requirements to be installed when just updating requirements.txt.

ocw/lib/azure.py
```diff
@@ -172,10 +177,38 @@ def cleanup_disks_from_rg(self) -> None:
...
+            vm.storage_profile.image_reference.id.lower()
+            for vm in self.compute_mgmt_client().virtual_machines.list_all()
+            if vm.storage_profile.image_reference.id is not None
+        }
+        active_ids = scale_set_gallery_ids | vm_gallery_ids
+
+        for item in self.list_galleries_by_resource_group(self.__resource_group):
+            if self.is_outdated(item.changed_time):
+                gallery_id = item.id.lower()
+                if any(i.startswith(gallery_id) for i in active_ids):
```
asmorodskyi, line 207: `active_ids` is dict , right ? so maybe something like `gallery_id in active_ids.keys()` :thinking:

ricardobranco777: > `active_ids` is dict , right ? so maybe something like `gallery_id in active_ids.keys()` 🤔

It's a set.

ricardobranco777: > ok than even simpler `gallery_id in active_ids` , why do we need `start_with` ? is there could be case that ID not fully match but we still don't want to treat it as the same ?

I with it were that simpler :)

Omitting the subscription in paths for `xxx`:

```
>>> client = ComputeManagementClient(credentials, subscription)
>>> mgmt_client = ResourceManagementClient(credentials, subscription)
>>> galleries = list(mgmt_client.resources.list_by_resource_group("openqa-upload", filter="resourceType eq 'Microsoft.Compute/galleries'", expand="changedTime"))
>>> galleries[0].id
'/subscriptions/xxx/resourceGroups/openqa-upload/providers/Microsoft.Compute/galleries/test_image_gallery'
>>> vms = list(client.virtual_machines.list_all())
>>> vm_gallery_ids = [vm.storage_profile.image_reference.id for vm in vms if vm.storage_profile.image_reference.id is not None]
>>> vm_gallery_ids[0]
'/subscriptions/xxx/resourceGroups/openqa-upload/providers/Microsoft.Compute/galleries/test_image_gallery/images/SLE-MICRO-5.2-BYOS-AZURE-X64-GEN2/versions/1.304.101'
```

So we have that `vm_gallery_ids[0]` starts with `galleries[0].id`.  Now I noticed that it's much safer to append a slash to `galleries[0].id` in the call to `startswith()`

pdostal:
The only issue is that we should not remove galleries but gallery versions.

ricardobranco777:
> I may not understand the code that well but I for example don't know what gallery id is.

An ID in Azure is the URL of the resource.

ocw/lib/azure.py
```diff
@@ -157,25 +163,41 @@ def cleanup_blob_containers(self) -> None:
...
                 else:
                     if self.dry_run:
                         self.log_info(f"Deletion of disk {item.name} skipped due to dry run mode")
                     else:
                         self.log_info(f"Delete disk '{item.name}'")
                         self.compute_mgmt_client().disks.begin_delete(self.__resource_group, item.name)
+
+    def cleanup_versions(self) -> None:
+        self.log_dbg("Call cleanup_versions")
+        for gallery in self.list_galleries():
```
grisu48, line 191: I would not do the version cleanup in all galleries. We only have one specified Image Gallery for our tests.
I would configure the image gallery in the configuration file, and perform the cleanup only on this one, so that we can have test galleries, which are not being affected. This was useful while developing and might come in handy, when we have to do some testing.

In addition I would introduce a check, that if the `pcw_ignore` tag is applied to a image gallery, then the whole gallery and it's definitions will be skipped.

ricardobranco777: > I would not do the version cleanup in all galleries. We only have one specified Image Gallery for our tests. I would configure the image gallery in the configuration file, and perform the cleanup only on this one, so that we can have test galleries, which are not being affected. This was useful while developing and might come in handy, when we have to do some testing.

Fixed.

> In addition I would introduce a check, that if the `pcw_ignore` tag is applied to a image gallery, then the whole gallery and it's definitions will be skipped.

I went further and introduced tag checking also for images and versions.

ocw/lib/azure.py
```diff
@@ -157,25 +161,52 @@ def cleanup_blob_containers(self) -> None:
...
-                    self.log_warn(f"Disk is in use - unable delete {item.name}")
+                    self.log_warn(f"Disk is in use - skipping {item.name}")
                 else:
                     if self.dry_run:
                         self.log_info(f"Deletion of disk {item.name} skipped due to dry run mode")
                     else:
                         self.log_info(f"Delete disk '{item.name}'")
                         self.compute_mgmt_client().disks.begin_delete(self.__resource_group, item.name)
+
+    def cleanup_versions(self) -> None:
```
asmorodskyi, line 187: this is too broad let's give more certain name 
```suggestion
    def cleanup_gallery_img_versions(self) -> None:
```

ocw/lib/azure.py
```diff
@@ -157,25 +161,52 @@ def cleanup_blob_containers(self) -> None:
...
+        self.log_dbg("Call cleanup_versions")
+        try:
+            gallery = self.compute_mgmt_client().galleries.get(self.__resource_group, self.__gallery)
+        except ResourceNotFoundError:
+            self.log_err(f"No such gallery in resource group {self.__resource_group}: {self.__gallery}")
+            return
```
asmorodskyi, line 193: this is something what not expected to happen so if it does I would like error to propagate further and cause mail send to us , so just remove this handler or replace return with raising same exception

ocw/lib/azure.py
```diff
@@ -157,25 +161,52 @@ def cleanup_blob_containers(self) -> None:
...
+        if Instance.TAG_IGNORE in gallery.tags:
+            self.log_err(f"Gallery in resource group {self.__resource_group} has {Instance.TAG_IGNORE} tag: {self.__gallery}")
+            return
+        for image in self.compute_mgmt_client().gallery_images.list_by_gallery(self.__resource_group, gallery.name):
+            if Instance.TAG_IGNORE in image.tags:
+                continue
```
asmorodskyi, line 199: let's insert info level log message , we still want to see amount of such images

ocw/lib/azure.py
```diff
@@ -157,25 +161,52 @@ def cleanup_blob_containers(self) -> None:
...
+            for version in self.compute_mgmt_client().gallery_image_versions.list_by_gallery_image(
+                    self.__resource_group, gallery.name, image.name):
+                if Instance.TAG_IGNORE in version.tags:
+                    continue
```
asmorodskyi, line 203: let's insert info level log message , we still want to see amount of such images

ocw/lib/azure.py
asmorodskyi: you misunderstood me , I didn't want to log every image exists in gallery but only one that has IGNORE tag on it

ocw/lib/azure.py
ricardobranco777: Fixed.

ocw/lib/azure.py
ricardobranco777: Fixed.

asmorodskyi (approved):
LGTM

grisu48 (approved):
LGTM! This approach should work, but we will need to see if it actually does once deploying. Could not identify any obvious mistakes or blockers, from my side it's a go.

:rocket:

PR #289, opened 2023-08-01
Fix issues with podman

ricardobranco777:
This PR:
- Suggests fixes for broken podman environments.
- Fixes exception handling and getting container port.
- Updates the documentation.

Tested on Ubuntu 23.10 & openSUSE Tumbleweed with latest podman, as well as Ubuntu VM in Github Actions with older podman shipped from Ubuntu repos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,22 +170,7 @@ pip install -r requirements_test.txt
 make test
 ```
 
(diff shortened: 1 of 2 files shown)


pdostal (approved):
I like it

PR #290, opened 2023-08-03
Cleanup old helm-test k8s namespaces

pdostal:
* Related ticket: [poo#123511](https://progress.opensuse.org/issues/123511)

PR #294, opened 2023-08-10
Fix pylint warnings in ocw/lib/k8s.py

pdostal:
* Follows #290

### DIFF
--- a/ocw/lib/aks.py
+++ b/ocw/lib/aks.py
@@ -53,4 +53,4 @@ class AKS(Provider):
         for cluster in clusters:
             self.log_info(f"Cleaning namespaces in AKS cluster {cluster['cluster_name']}")
             client = self.kubectl_client(cluster["resource_group"], cluster["cluster_name"]).CoreV1Api()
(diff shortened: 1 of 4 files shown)


ricardobranco777 (dismissed):
LGTM.

We should always check pylint locally, but I think we can improve our workflow by having pylint in CI not tolerate any errors.  But for this we'd have to add explicit `disable`'s in the code. WDYT?

ricardobranco777 (approved):

pdostal:
> We should always check pylint locally, but I think we can improve our workflow by having pylint in CI not tolerate any errors. But for this we'd have to add explicit `disable`'s in the code. WDYT?

That's rather question for @asmorodskyi .

I'd anyways continue in separate pull request.

asmorodskyi:
> LGTM.
> 
> We should always check pylint locally, but I think we can improve our workflow by having pylint in CI not tolerate any errors. But for this we'd have to add explicit `disable`'s in the code. WDYT?

LGTM for decrease tolerance here , main thing that we not merge with failed pipeline , until we keep this I am fine :)

PR #297, opened 2023-08-14
Install SUSE certs and enable SSL warnings

ricardobranco777:
Install SUSE certs and enable SSL warnings

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,8 @@ COPY requirements.txt /pcw/
 # * Install system requirements
 # * Install pip requirements
 # * Empty system cache to conserve some space
-RUN zypper -n in gcc libffi-devel && pip install --no-cache-dir -r /pcw/requirements.txt && zypper clean && rm -rf /var/cache
+RUN source /etc/os-release && zypper addrepo -G -cf "https://download.opensuse.org/repositories/SUSE:/CA/$VERSION_ID/SUSE:CA.repo" && \
+    zypper -n in ca-certificates-suse gcc libffi-devel && pip install --no-cache-dir -r /pcw/requirements.txt && zypper clean && rm -rf /var/cache
 
 # Copy program files only
 COPY ocw  /pcw/ocw/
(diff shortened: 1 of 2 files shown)


pdostal (dismissed):

ocw/lib/openqa.py
```diff
@@ -26,7 +18,7 @@ def get_url(server):
             got = openqa_client.client.requests.head(
                 url,
                 timeout=5,
-                verify=bool(os.environ.get('REQUESTS_CA_BUNDLE')),
+                verify=False,
```
dcermak, line 21: You meant this, right?
```suggestion
                verify=True,
```

ocw/lib/openqa.py
```diff
@@ -48,7 +40,7 @@ def __new__(cls, **kwargs):
     def __init__(self, **kwargs):
         kwargs.pop("server")
         self.__client = openqa_client.client.OpenQA_Client(server=self.server, **kwargs)
-        self.__client.session.verify = bool(os.environ.get('REQUESTS_CA_BUNDLE'))
+        self.__client.session.verify = False
```
clanig, line 43: This is disabling TLS verification as well.

ricardobranco777: > This is disabling TLS verification as well.

We can't verify every openQA instance with self-signed certs.

ocw/lib/openqa.py
grisu48: I was also puzzled by this but this is valid.

The reason is that `pcw` needs to query any openQA instances for job status updates. This includes openQA development instances, where many people use self-signed certificates.

grisu48 (approved):

ocw/lib/openqa.py
clanig: OK for me as well then. We might find a better overall solution to everything later.
However, I would appreciate to have a short comment why TLS verification is disabled whenever this is done in any code. @ricardobranco777 I would appreciate very much if you could add a short comment to the sections.

ocw/lib/openqa.py
clanig: OK, I see you are right. Thanks for the explanation!

clanig:
Thank you for the changes. Looks great from my perspective.

Dockerfile
```diff
@@ -28,5 +29,7 @@ VOLUME /pcw/db
 
 EXPOSE 8000/tcp
 
+ENV REQUESTS_CA_BUNDLE=/etc/ssl/ca-bundle.pem
```
grisu48, line 32:
```suggestion
# Required to use system certs in python-requests
ENV REQUESTS_CA_BUNDLE=/etc/ssl/ca-bundle.pem
```

grisu48 (approved):

asmorodskyi:
I think we also need to tweak Dockerfile_dev so we won't see this warnings during developing new features

PR #298, opened 2023-08-15
gce: Fix exception in list_zones

ricardobranco777:
Fix exception in `list_zones()` that's preventing pcw from cleaning up due to region `europe-west10` not existing.

- Related ticket: https://progress.opensuse.org/issues/134267

### DIFF
--- a/ocw/lib/gce.py
+++ b/ocw/lib/gce.py
@@ -91,13 +91,17 @@ class GCE(Provider):
         return [region["name"] for region in regions]
 
     def list_zones(self, region) -> list:
-        region = (
-            self.compute_client()
-            .regions()
-            .get(project=self.project, region=region)
-            .execute()
-        )
-        return [basename(z) for z in region["zones"]]
+        try:
+            region = (
+                self.compute_client()
+                .regions()
+                .get(project=self.project, region=region)
+                .execute()
+            )
+            return [basename(z) for z in region["zones"]]
+        except (KeyError, HttpError) as exc:
+            self.log_dbg("list_zones: %s", exc)
+            return []
 
     def delete_instance(self, instance_id, zone) -> None:
         self._delete_resource(


ocw/lib/gce.py
```diff
@@ -91,13 +91,17 @@ def list_regions(self) -> list:
...
     def list_zones(self, region) -> list:
-        region = (
-            self.compute_client()
-            .regions()
-            .get(project=self.project, region=region)
-            .execute()
-        )
-        return [basename(z) for z in region["zones"]]
+        try:
+            region = (
```
grisu48, line 95: Why not getting the regions and check, if the desired region is in the list? This error message looks like spam to me.

ricardobranco777: > Why not getting the regions and check, if the desired region is in the list? This error message looks like spam to me.

This is dynamic as regions and zones can suddenly "disappear".  That's why they can't be cached.  Also, the code that calls `list_zones()` iterates over `list_regions()`.  We wouldn't want to call `list_regions()` twice.

pdostal (approved):

ocw/lib/gce.py
grisu48: ok

ocw/lib/gce.py
```diff
@@ -91,13 +91,17 @@ def list_regions(self) -> list:
...
+                self.compute_client()
+                .regions()
+                .get(project=self.project, region=region)
+                .execute()
+            )
+            return [basename(z) for z in region["zones"]]
+        except (KeyError, HttpError) as exc:
```
asmorodskyi, line 102: again the same as in #300  we should not mute such broad exceptions . I assume you trying to handle properly case when google ( very often ) returning regions in list_regions which later in other calls returning "region does not exists " ? generally I agree with this but we must be more precise . what you catching here is too broad ....

ricardobranco777: This is better than the situation we  had before where the exception prevented the cleanup from going on.  We still must catch it, and at least we're doing something with it like logging the error.

asmorodskyi: it is big question to if it is better now . You catching all KeyError/HttpError and only doing debug message about that so they currently FULLY muted in prod instance where we have INFO log level so if something is broken in this call we missing some part of regions from cleanup and NOT getting any notifications about that anywhere I am not sure if it can be called better

ricardobranco777: This was fixed in a subsequent PR which changed log_dbg to log_err.

asmorodskyi: thanks !

PR #299, opened 2023-08-15
Verify TLS by default to avoid warnings

ricardobranco777:
Verify TLS by default to avoid SSLError warnings when connecting to openQA.  These warnings can't be disabled in `requests` for openQA only because doing so would also disable them when connecting to cloud providers.

### DIFF
--- a/ocw/lib/openqa.py
+++ b/ocw/lib/openqa.py
@@ -1,18 +1,25 @@
 from urllib.parse import urlparse
 from cachetools import cached
-from requests.exceptions import RequestException
+import requests
+from requests.exceptions import RequestException, SSLError
 import openqa_client.client
 from openqa_client.const import JOB_STATE_CANCELLED
 from openqa_client.exceptions import OpenQAClientError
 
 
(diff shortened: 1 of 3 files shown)


grisu48 (approved):
If the approach is "let's use TLS whenever possible and insecure TLS when needed", then this is good.

LGTM!

ocw/lib/openqa.py
```diff
@@ -1,29 +1,36 @@
...
+from requests.exceptions import RequestException, SSLError
 import openqa_client.client
 from openqa_client.const import JOB_STATE_CANCELLED
 from openqa_client.exceptions import OpenQAClientError
 
 
-# We don't verify TLS server certificates because we
-# may encounter self-signed or expired certificates
-DEFAULT_VERIFY = False
+DEFAULT_TIMEOUT = 3
```
asmorodskyi, line 10: why you decreased timeout ?

ricardobranco777: I felt that having the same timeout at the function that verifies TLS and the connection as way too much.

PR #301, opened 2023-08-22
Include ignore field in instances.json

grisu48:
The instances.json path should also include the ignore field to allow external applications to hide ignored instances.


pdostal (approved):

ricardobranco777 (approved):
LGTM

PR #304, opened 2023-08-31
More fixes

ricardobranco777:
This PR:
- Disables `insecure=True` in Openstack now that we install SUSE certificates.
- Fix a couple of minor pylint issues.

### DIFF
--- a/ocw/lib/openstack.py
+++ b/ocw/lib/openstack.py
@@ -25,7 +25,6 @@ class Openstack(Provider):
         if self.__client is None:
             self.__client = openstack.connect(
                 debug=bool(DEBUG),
-                insecure=True,  # Trust the certificate
                 auth_url=self.get_data('auth_url'),
                 project_name=self.get_data('project_name'),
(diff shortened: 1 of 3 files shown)


ocw/lib/openstack.py
```diff
@@ -25,7 +25,7 @@ def client(self) -> None:
         if self.__client is None:
             self.__client = openstack.connect(
                 debug=bool(DEBUG),
-                insecure=True,  # Trust the certificate
+                insecure=False,
```
asmorodskyi, line 28: I am pretty sure that `insecure=False` is default so you can just drop the line

asmorodskyi (dismissed):
LGTM

ocw/lib/openstack.py
ricardobranco777: > I am pretty sure that `insecure=False` is default so you can just drop the line

Fixed.

asmorodskyi (approved):
LGTM

pdostal (approved):

PR #311, opened 2023-09-01
Temporary fix for openQA-python-client

ricardobranco777:
Temporary fix for openQA-python-client


ricardobranco777:
Let's wait for https://github.com/os-autoinst/openQA-python-client/pull/47

PR #313, opened 2023-09-04
Prefer http over https for OpenQA

ricardobranco777:
As we're only getting job statuses without using credentials we can prefer HTTP over HTTPS.  This avoids the warnings in the case of expired or self-signed certificates.

For the case of legitimate certificates signed by SUSE, we wait for:
https://github.com/os-autoinst/openQA-python-client/pull/47

### DIFF
--- a/ocw/lib/openqa.py
+++ b/ocw/lib/openqa.py
@@ -27,7 +27,7 @@ def get_url(server: str) -> str:
         server = server.rstrip('/').replace("_", ".")
     if urlparse(server).scheme != "":
         return server
-    for scheme in ("https", "http"):
+    for scheme in ("http", "https"):
         try:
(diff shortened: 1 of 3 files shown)


asmorodskyi (dismissed):
LGTM

asmorodskyi:
can you please disable coverage checks as we agreed ?but preferable to keep pasting of cov report into each PR if it is possible

ricardobranco777:
> can you please disable coverage checks as we agreed ?but preferable to keep pasting of cov report into each PR if it is possible

There's no easy or obvious way to do it.  I think the best compromise is to just ignore it ourselves when merging.

asmorodskyi (approved):
LGTM

PR #315, opened 2023-09-05
Add codecov.yml allowing to make codecov check informational

asmorodskyi:
In majority of PRs which we merging codecov checks are red but we merging them anyway , hence there is no point to make PRs red with reasons which we tend to ignore


ricardobranco777 (approved):
Wow.  It was harder than I imagined.

LGTM

PR #328, opened 2023-11-30
GCE: Add 'tf-network' to the list of protected networks

pdostal:
Instead of creating networking resources again and again in each openQA run we now have single main network (`tf-network` see [gitlab.suse.de/qac/terraform](https://gitlab.suse.de/qac/terraform) for more details) which we reuse. This pull request attempts to protect this network and its subnetworks and firewalls from the great Public Cloud Watcher in Google Cloud Platform.


ocw/lib/gce.py
```diff
@@ -11,7 +12,11 @@
 
 class GCE(Provider):
     __instances = {}
-    __skip_networks = frozenset({"default"})
+
+    if PCWConfig.has('cleanup/gce-skip-networks'):
```
asmorodskyi, line 16: `__skip_networks = frozenset(ConfigFile().getList('cleanup/gce-skip-networks', "default"))`  - will replace whole if/else block

ocw/lib/gce.py
```diff
@@ -11,7 +12,11 @@
...
+        __skip_networks = frozenset(ConfigFile().getList('cleanup/gce-skip-networks'))
```
asmorodskyi, line 17: can you elaborate about frozenset ? why do we need it ?

ocw/lib/gce.py
pdostal: No. The `.getList()` does not have 'default value'.

asmorodskyi: https://github.com/SUSE/pcw/blob/master/webui/PCWConfig.py#L47

```
def getList(self, config_path: str, default: list = None) -> list:
        if default is None:
            default = []
        return [i.strip() for i in self.get(config_path, ','.join(default)).split(',')]
```

and what about second parameter named `default` ?

ocw/lib/gce.py
asmorodskyi: ah sorry I didn't notice that it was there before you , topic closed

pdostal:
Deployed in https://gitlab.suse.de/qac/publiccloud-qa-suse-de/-/merge_requests/132

PR #329, opened 2023-11-30
Return response on update_status in all cases

grisu48:
Always return a reply when update/status is queried.

Redirecting the `/update/status` to `/instances` in the case that a certain http header is not set is confusing and should not happen. The API path should always return something valid or at least throw an error if a input parameter is missing.


ricardobranco777 (approved):
LGTM. Please fix the indentation within the block.

PR #336, opened 2023-12-21
Create dedicated command to collect public cloud stats

asmorodskyi:


ocw/management/commands/dumpstate.py
```diff
@@ -0,0 +1,10 @@
+from django.core.management.base import BaseCommand
+from ocw.lib.dump_state import dump_state
+
+
+class Command(BaseCommand):
```
ricardobranco777, line 5: I would use a less generic name for this class.

ocw/lib/influx.py
```diff
@@ -53,5 +43,10 @@ def write(self, measurement: str, field: str, value: int) -> None:
             point = Point(measurement).field(field, value)
             try:
                 self.__client.write(bucket=self.bucket, org=self.org, record=point)
-            except (InfluxDBError, HTTPError, TimeoutError) as exception:
-                logger.warning(f"Failed to write to influxdb(record={point}): {exception}")
+            except (InfluxDBError, HTTPError) as exception:
+                logger.warning("Failed to write to influxdb(record=%s): %s", point, exception)
+
+    def dump_resource(self, provider: str, field: str, dump_method: Callable):
```
ricardobranco777, line 49:
```suggestion
    def dump_resource(self, provider: str, field: str, dump_method: Callable) -> None:
```

ricardobranco777 (dismissed):
Couple of suggestions. LGTM.

ocw/management/commands/dumpstate.py
asmorodskyi: please check ALL other files in this folder :) I am not 100% sure but this is might be requirement from django how this "commands flow" is working . In any case no one except django itself suppose to create instances of this class so this one and only place where this name will be used

ocw/lib/influx.py
asmorodskyi: make sense

PR #345, opened 2024-01-29
VPC cleanup: use DestinationIpv6CidrBlock when available

asmorodskyi:


requirements.txt
```diff
@@ -1,9 +1,9 @@
 boto3
-azure-mgmt-compute==30.1.0
+azure-mgmt-compute==30.4.0
 azure-mgmt-storage==21.1.0
-azure-mgmt-resource==23.0.1
-azure-storage-blob==12.17.0
-azure-identity==1.14.0
+azure-mgmt-resource==23.1.0b2
```
ricardobranco777, line 4: Are you sure you want to include a beta version?

requirements_k8s.txt
```diff
@@ -1,7 +1,8 @@
 oauth2client
 kubernetes
+azure-cli==2.56.0
 google-api-python-client==2.96.0
-azure-mgmt-containerservice==22.0.0
-azure-identity==1.14.0
-azure-mgmt-resource==23.0.1
+azure-mgmt-containerservice==28.0.0
+azure-identity==1.14.1
+azure-mgmt-resource==23.1.0b2
```
ricardobranco777, line 7: Same as above.

ricardobranco777 (approved):
Couple of suggestions. Otherwise LGTM.

pdostal (approved):
Maybe at least mention that Azure issue you're dealing with here as well.

requirements.txt
asmorodskyi: I gave you extend answer in PM . Here I will give short answer - it is choice of MS developers

requirements_k8s.txt
asmorodskyi: I will unconditionally close this thread to keep discussions regarding this topic in single place

asmorodskyi:
> Maybe at least mention that Azure issue you're dealing with here as well.

PR description updated please check if it is ok now

PR #358, opened 2024-03-07
Start counting image versions for Azure

asmorodskyi:
Currently existing counters ( image and blobs ) always returning 0s because we actually storing images currently as image versions so we need to add counter for this

### DIFF
--- a/ocw/lib/azure.py
+++ b/ocw/lib/azure.py
@@ -18,9 +18,9 @@ class Azure(Provider):
 
     def __init__(self, namespace: str):
         super().__init__(namespace)
-        self.__resource_group = PCWConfig.get_feature_property('cleanup', 'azure-storage-resourcegroup', namespace)
+        self.__resource_group: str = str(PCWConfig.get_feature_property('cleanup', 'azure-storage-resourcegroup', namespace))
         self.check_credentials()
-        self.__gallery = PCWConfig.get_feature_property('cleanup', 'azure-gallery-name', namespace)
+        self.__gallery: str = str(PCWConfig.get_feature_property('cleanup', 'azure-gallery-name', namespace))
 
     def __new__(cls, namespace: str) -> 'Azure':
         if namespace not in Azure.__instances:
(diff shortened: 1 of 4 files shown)


ocw/lib/azure.py
```diff
@@ -206,3 +206,14 @@ def cleanup_gallery_img_versions(self) -> None:
...
+
+    def get_img_versions_count(self) -> int:
+        self.log_dbg("Call get_img_versions_count")
+        gallery = self.compute_mgmt_client().galleries.get(self.__resource_group, self.__gallery)
+        all_img_versions = 0
+        for image in self.compute_mgmt_client().gallery_images.list_by_gallery(self.__resource_group, gallery.name):
+            img_version = len(list(self.compute_mgmt_client().gallery_image_versions.list_by_gallery_image(
+                    self.__resource_group, gallery.name, image.name)))
+            self.log_dbg(f"{image.name} has {img_version} versions")
+            all_img_versions+=img_version
```
ricardobranco777, line 218:
```suggestion
            all_img_versions += img_version
```

ocw/lib/azure.py
```diff
@@ -206,3 +206,14 @@ def cleanup_gallery_img_versions(self) -> None:
                         self.compute_mgmt_client().gallery_image_versions.begin_delete(
                                 self.__resource_group, gallery.name, image.name, version.name
                         )
+
+    def get_img_versions_count(self) -> int:
+        self.log_dbg("Call get_img_versions_count")
+        gallery = self.compute_mgmt_client().galleries.get(self.__resource_group, self.__gallery)
+        all_img_versions = 0
+        for image in self.compute_mgmt_client().gallery_images.list_by_gallery(self.__resource_group, gallery.name):
+            img_version = len(list(self.compute_mgmt_client().gallery_image_versions.list_by_gallery_image(
```
ricardobranco777, line 215: I would rename this variable because right now it seems like holding a version, instead of the number of image versions.

```suggestion
            img_versions = len(list(self.compute_mgmt_client().gallery_image_versions.list_by_gallery_image(
```

ricardobranco777 (dismissed):
Some suggestions. LGTM

ricardobranco777 (approved):

PR #371, opened 2024-05-21
Fix selenium test

ricardobranco777:
This PR fixes Selenium test for PCW:

- Updates Selenium version.
- Builds base image before building pcw itself.
- Adds a sleep before creating user in database to allow for the migrations at startup to run.
- Tidies up the Xpaths for the login & logout buttons.

### DIFF
--- a/tests/test_webui.py
+++ b/tests/test_webui.py
@@ -4,6 +4,8 @@ import random
 import os
 import shutil
 import sys
+import time
+from enum import StrEnum
 from subprocess import DEVNULL
 
 import pytest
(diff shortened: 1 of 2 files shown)


mimi1vx (approved):

tests/test_webui.py
```diff
@@ -44,22 +45,21 @@ def client():
...
 
-@pytest.fixture(scope="session")
+@pytest.fixture
 def random_port():
     # Get random number for ephemeral port, container and image name
     # Typical values from /proc/sys/net/ipv4/ip_local_port_range
     return random.randint(32768, 60999)
 
 
-@pytest.fixture(scope="session")
```
mimi1vx, line 54: why remove session scope?

ricardobranco777: > why remove session scope?

I'm not sure why I added it in the first place, but the `tmp_path` fixture refused to work with it enabled.  The default scope  is ok since we're not reusing the fixtures defined in this test.

asmorodskyi (approved):
LGTM

PR #373, opened 2024-05-24
Enable stat collection for EC2

asmorodskyi:


asmorodskyi:
I messed up git tree here closing same changes will be in #380

PR #376, opened 2024-05-24
Drop base images

ricardobranco777:
Drop base images as it creates an issue where dependencies introduced via requirements.txt are not properly tested in CI.


pdostal (approved):

asmorodskyi:
just for the reference PR where base image was introduced https://github.com/SUSE/pcw/pull/350/files

PR #378, opened 2024-05-29
gce: Cleanup buckets

ricardobranco777:
Cleanup GCE buckets.

Related MR: https://gitlab.suse.de/qac/publiccloud-qa-suse-de/-/merge_requests/159

### DIFF
--- a/ocw/lib/gce.py
+++ b/ocw/lib/gce.py
@@ -6,7 +6,7 @@ from dateutil.parser import parse
 import googleapiclient.discovery
 from googleapiclient.errors import HttpError
 from google.oauth2 import service_account
-from webui.PCWConfig import ConfigFile
+from webui.PCWConfig import ConfigFile, PCWConfig
 from .provider import Provider
 
 
@@ -21,9 +21,14 @@ class GCE(Provider):
     def __init__(self, namespace):
         super().__init__(namespace)
 
+        try:
+            self.__bucket = PCWConfig.get_feature_property('cleanup', 'gce-bucket', namespace)
+        except LookupError:
+            self.__bucket = None
         self.__skip_networks = frozenset(ConfigFile().getList('cleanup/gce-skip-networks', ["default"]))
 
         self.__compute_client = None
+        self.__storage_client = None
         self.private_key_data = self.get_data()
         self.project = self.private_key_data["project_id"]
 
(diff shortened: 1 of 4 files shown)


webui/PCWConfig.py
```diff
@@ -55,11 +55,12 @@ class PCWConfig():
     @staticmethod
     def get_feature_property(feature: str, feature_property: str, namespace: str | None = None) -> str | int:
         default_values: dict[str, dict[str, int | type[int] | str | type[str] | type[str] | None]] = {
-            'cleanup/max-age-hours': {'default': 24 * 7, 'return_type': int},
             'cleanup/azure-gallery-name': {'default': 'test_image_gallery', 'return_type': str},
             'cleanup/azure-storage-resourcegroup': {'default': 'openqa-upload', 'return_type': str},
             'cleanup/azure-storage-account-name': {'default': 'openqa', 'return_type': str},
             'cleanup/ec2-max-age-days': {'default': -1, 'return_type': int},
+            'cleanup/gce-bucket': {'default': 'test', 'return_type': str},
```
asmorodskyi, line 62:
```suggestion
            'cleanup/gce-bucket': {'default': None, 'return_type': str},
```

asmorodskyi:
can you please also add tests for cleanup_blobs function ?

webui/PCWConfig.py
ricardobranco777: Fixed.

ocw/lib/gce.py
```diff
@@ -21,9 +21,11 @@ def __new__(cls, namespace):
     def __init__(self, namespace):
         super().__init__(namespace)
 
+        self.__bucket = PCWConfig.get_feature_property('cleanup', 'gce-bucket', namespace)
```
asmorodskyi, line 24: we also need to support case when there is pcw.ini where this variable is NOT defined .

ocw/lib/gce.py
```diff
@@ -142,6 +154,16 @@ def cleanup_all(self) -> None:
         self.cleanup_subnetworks()
         self.cleanup_networks()
 
+    def cleanup_blobs(self) -> None:
```
asmorodskyi, line 157: I would add here check if bucket is defined and if not do early exit

ricardobranco777: Fixed in the `cleanup_all` method.  Otherwise I would have to make the `__bucket` attribute non-private to properly mock it in the tests.

ocw/lib/gce.py
ricardobranco777: Fixed.

asmorodskyi (approved):
LGTM

PR #388, opened 2024-06-24
Do not remove EC2 images with 'pcw_ignore' tag set to '1'

pdostal:
* Related ticket: [poo#162770](https://progress.opensuse.org/issues/162770)


ocw/lib/ec2.py
```diff
@@ -352,8 +352,13 @@ def cleanup_images(self, valid_period_days: float) -> None:
             self.log_dbg(f"Found {len(response['Images'])} images in {region}")
             for img in response['Images']:
                 if EC2.is_outdated(parse(img['CreationDate']), valid_period_days):
-                    if self.dry_run:
-                        self.log_info(f"Image deletion {img['ImageId']} skipped due to dry run mode")
+                    tags = img['Tags'] if 'Tags' in img else []
+                    pcw_ignore_tag = next((tag for tag in tags if tag['Key'] == 'pcw_ignore' and tag['Value'] == '1'), None)
```
ricardobranco777, line 356: Let's consider any non-zero value as truthy and return False otherwise:

```suggestion
                    pcw_ignore_tag = next((tag for tag in tags if tag['Key'] == 'pcw_ignore' and tag['Value'] != '0'), False)
```

ocw/lib/ec2.py
```diff
@@ -352,8 +352,13 @@ def cleanup_images(self, valid_period_days: float) -> None:
             self.log_dbg(f"Found {len(response['Images'])} images in {region}")
             for img in response['Images']:
                 if EC2.is_outdated(parse(img['CreationDate']), valid_period_days):
-                    if self.dry_run:
-                        self.log_info(f"Image deletion {img['ImageId']} skipped due to dry run mode")
+                    tags = img['Tags'] if 'Tags' in img else []
```
ricardobranco777, line 355: As this is a dict:

```suggestion
                    tags = img.get('Tags', [])
```

ricardobranco777 (approved):
Couple of suggestions. LGTM.

ocw/lib/ec2.py
pdostal: Can I do following?
```suggestion
                    pcw_ignore_tag = next((tag for tag in tags if tag['Key'] == Instance.TAG_IGNORE), None)
```
as:
1) I see it in other places in the code
2) I have the impression that that's what @asmorodskyi wants :smile:

ricardobranco777: > Can I do following?
> 
> as:
> 
> 1. I see it in other places in the code

Yes. Much better.

ricardobranco777 (approved):

asmorodskyi:
LGTM

PR #397, opened 2024-07-23
Add dump stat for GCE

asmorodskyi:


ocw/lib/gce.py
```diff
@@ -264,3 +264,22 @@ def cleanup_networks(self) -> None:
...
                     self.compute_client().networks, network["name"], project=self.project, network=network["name"]
                 )
+
+    def count_all_instances(self) -> int:
+        return len(self.list_all_instances())
+
+    def count_all_images(self) -> int:
+        return len(self._paginated(self.compute_client().images, project=self.project))
+
+    def count_all_disks(self) -> int:
```
ricardobranco777, line 274: We could use a list comprehension here but I leave it to you.

ricardobranco777 (approved):

m-dati (approved):

ocw/lib/gce.py
asmorodskyi: IMHO because loop is nested it would look a bit messy . So if you don't mind I will keep it as is

PR #403, opened 2024-08-29
Remove support for Openstack

ricardobranco777:
Remove support for Openstack.

Perhaps missing in this PR is modifying this line:
https://github.com/SUSE/pcw/blob/master/ocw/migrations/0011_alter_instance_provider.py#L16

PR #404, opened 2024-09-16
Mute common error when deleting VPC

asmorodskyi:
This PR contains 3 things : 
- remove flow which was sending emails about VMs which running too long 
- Tweak DB Schema with remove of Openstack flow ( was missing in #403  ) 
- mute most common error which happens on VPC deletion


m-dati (approved):

ocw/lib/ec2.py
```diff
@@ -325,7 +325,12 @@ def vpc_can_be_deleted(self, resource_vpc, vpc_id) -> bool:
 
     def report_cleanup_results(self, vpc_errors: list, vpc_notify: list, vpc_locked: list) -> None:
         if len(vpc_errors) > 0:
```
ricardobranco777, line 327: This `if` is redundant with the block below.

ricardobranco777 (approved):
Minor suggestion. LGTM.

ocw/lib/ec2.py
asmorodskyi: agree with you but unfortunately  I pressed merge button just before your comment popup :disappointed:

PR #425, opened 2025-01-30
Remove duplicate test

ricardobranco777:
`make test` runs the same `pytest --cov` that the codecov job runs as shown [here](https://github.com/SUSE/pcw/blob/master/Makefile#L23)

It just wastes CPU cycles and increases our carbon footprint for nothing.

Either we remove this or add `SKIP_SELENIUM=1` to one of them, as this test takes a lot of time building a Docker image and then running Firefox for Selenium.


asmorodskyi (approved):
LGTM

PR #432, opened 2025-02-27
Update requirements.txt

ricardobranco777:
Update requirements.txt

This updates all requirements to match versions fixed for Python 3.13, so we can update our base images from 3.11.


asmorodskyi (changes requested):
let's clarify about case sensitive package names . Also I want to make sure that proper testing is done

asmorodskyi (approved):
LGTM

PR #444, opened 2025-04-24
Fix deleting reset

volodymyrkatkalov:
https://progress.opensuse.org/issues/180635


ocw/lib/utils.py
```diff
@@ -0,0 +1,9 @@
+def parse_boolean(v) -> bool:
```
asmorodskyi, line 1: we can fallback to ANY value

asmorodskyi (dismissed):
LGTM but please fix CI

asmorodskyi (approved):
still LGTM :)

PR #458, opened 2025-07-01
Update dependencies

ricardobranco777:
Update dependencies.

Also drop the **build-essential** package from the pytest Github Actions workflow so it doesn't compile **uwsgi** in the Ubuntu worker but downloads the wheel instead.  Compiling fails due to missing openSSL 3 headers.  So instead of adding yet another package (libssl-devel), use wheels for everything and CI is even faster.


asmorodskyi (approved):
LGTM

PR #477, opened 2025-09-25
Improve exception handling on resource deletion in GCE

asmorodskyi:
1. Ease requirements for muting exceptions on Route deletion
2. Add new case necessary for instanceGroups cleanup

### DIFF
--- a/tests/test_gce.py
+++ b/tests/test_gce.py
@@ -241,6 +241,12 @@ def test_cleanup_routes_delete_default_route_not_raise_exception(gce_dry_run_fal
     gce_dry_run_false.cleanup_routes()
 
 
+def test_cleanup_instance_groups_delete_managed_group_not_raise_exception(gce_dry_run_false, mocked_resource):
+    setattr(gce_dry_run_false.compute_client, "instanceGroups", mocked_resource)
+    mocked_resource.error_reason = "resourceIsManaged"
+    gce_dry_run_false.cleanup_instance_groups()
+
+
 @mark.parametrize("dry_run", [True, False])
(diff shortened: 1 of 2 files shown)


tests/test_gce.py
```diff
@@ -240,6 +240,12 @@ def test_cleanup_routes_delete_default_route_not_raise_exception(gce_dry_run_fal
     mocked_resource.error_reason = mocked_resource_reason
     gce_dry_run_false.cleanup_routes()
 
+@mark.parametrize("mocked_resource_reason", ["resourceInUseByAnotherResource", "resourceIsManaged"])
+def test_cleanup_instance_groups_delete_managed_group_not_raise_exception(gce_dry_run_false, mocked_resource, mocked_resource_reason):
```
ricardobranco777, line 244: We're not raising an exception.

Also, there's no concept of managed groups in GCE, only instance groups.

I think this test is not needed.  A proper check would use the `caplog` fixture from pytest to check this.

asmorodskyi: > We're not raising an exception.
I don't understand what do you mean by this sentence ? I also not raising exception 

> I think this test is not needed.

this test is needed to cover this code : https://github.com/SUSE/pcw/pull/477/files#diff-0dfc4095c73002d356e289d8edf872dcb8a6624ba9adc082c0c331aa9789bb9cR98 :

```
   elif error_reason == 'resourceIsManaged' and resource_type.title() == 'Instancegroup':
                self.log_err(f"{resource_type.title()} '{resource_name}' can not be deleted. Error : {err}")
```

ricardobranco777 (dismissed):

grisu48 (dismissed):

